PR #3, opened 2018-12-01
Introducing syncable module

flbulgarelli:
It normalices the interface of resource classes, removes duplicated
logic and introduces `locate` message.

This change does not require modifications to `Mumukit::Sync` :sweat_smile:


luchotc:
Nice! Only one doubt: where is syncable getting RID from?

flbulgarelli:
> Nice! Only one doubt: where is syncable getting RID from?

Nice catch! It was a bug, that constant there is a nonsense. I have just fixed it and added some tests

PR #5, opened 2018-12-10
Feature course invitation relationship

flbulgarelli:
Introducing a classroom-like relationship between courses and invitations, and migrating the `invite!` behavior to this gem. 

This change requires a non-backward compatible migration on invitations.

### DIFF
--- a/app/models/invitation.rb
+++ b/app/models/invitation.rb
@@ -1,23 +1,32 @@
+# invitation:
+#   course
+#   code
+#   expiration_date
+
 class Invitation < ApplicationRecord
-  def self.find_by_code(code)
-    Invitation
-      .where(code: code)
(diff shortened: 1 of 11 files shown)


app/models/invitation.rb
```diff
@@ -28,9 +36,30 @@ def to_param
...
+
+  def notify!
+    Mumukit::Nuntius.notify_event! 'InvitationChanged', invitation: to_resource_h
+  end
+
   private
 
   def course_name
-    course.to_mumukit_slug.course
+    course.name
```
julian-berbel: Courses don't actually have a name, so I guess we should copy the previous logic to Course? as it is now, it breaks.

flbulgarelli: You are right. Classroom calls it name, but domain calls it code. We should rename or alias it.

PR #10, opened 2018-12-26
Feature replace edition mode

flbulgarelli:
Alternate PR for historic reasons only. Do not merge. See #9


flbulgarelli:
Side notice: a problem with this PR is that it heavily relies on `patch`, which is not the most efficient method.

flbulgarelli:
Related to https://github.com/mumuki/mumukit-core/pull/11

PR #19, opened 2019-02-14
Normalize slugs - Fixes mumuki/mumuki-bibliotheca-api#86

rodri042:
Fixes https://github.com/mumuki/mumuki-bibliotheca-api/issues/86

:exclamation:  Remember to release & update the `mumuki-domain` gem in `mumuki-bibliotheca-api` after merging.

### DIFF
--- a/app/models/concerns/with_slug.rb
+++ b/app/models/concerns/with_slug.rb
@@ -5,11 +5,16 @@ module WithSlug
   included do
     validates_presence_of :slug
     validates_uniqueness_of :slug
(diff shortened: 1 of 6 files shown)


flbulgarelli:
Nice! 

Could we normalize all content - guides, topics and books - instead of just guides? Perhaps that could be done in the abstract `Content` class or the `WithSlug` concern

rodri042:
@flbulgarelli I've moved the code to the `WithSlug` concern

PR #20, opened 2019-02-18
Fix add uid presence validation

julian-berbel:


app/models/user.rb
```diff
@@ -32,6 +32,7 @@ class User < ApplicationRecord
   after_initialize :init
 
   before_validation :set_uid!
+  validates :name, presence: true
```
flbulgarelli: oups :stuck_out_tongue:

PR #21, opened 2019-03-06
Feature orphan records

flbulgarelli:
Avoiding cascades, which although doesn't actually permit restoring information, it limits the effects of erroneous deletes.

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -9,8 +9,7 @@ class Assignment < ApplicationRecord
   has_many :messages,
            -> { where.not(submission_id: nil).order(date: :desc) },
            foreign_key: :submission_id,
-           primary_key: :submission_id,
-           dependent: :nullify
+           primary_key: :submission_id
 
   belongs_to :submitter, class_name: 'User'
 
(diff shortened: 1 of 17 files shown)


app/models/assignment.rb
```diff
@@ -6,7 +6,10 @@ class Assignment < ApplicationRecord
 
   belongs_to :exercise
   has_one :guide, through: :exercise
-  has_many :messages, -> { where.not(submission_id: nil).order(date: :desc) }, foreign_key: :submission_id, primary_key: :submission_id, dependent: :delete_all
```
julian-berbel: I'm not sure we can actually remove this delete 🤔. I believe we added it explicitly because not deleting them was causing trouble. https://github.com/mumuki/mumuki-laboratory/issues/1179

flbulgarelli: Nice catch! So please revert that single deletion and merge

julian-berbel:
I did a few more tests on this.

Deleting discussions (without their messages) on destroying exercise seems to be enough to keep user profile from breaking, though I don't know if keeping the messages without the discussion itself makes sense.

Also, if we remove the exercise -> assignment dependency, if a user sends a classroom message and the exercise is deleted, this will cause the profile to break too. Should we be deleting classroom messages when deleting an exercise too? (note the messages are associated to the assignment, not the exercise directly)

PR #26, opened 2019-04-02
Fixing error and raising GoneError instead

flbulgarelli:
When user accesses an expired invitation, it raise a 500, since the exception namespace for `NotFoundError` was missing. 

However raising a `GoneError` has a lot more sense here, so I also changed that


flbulgarelli:
Nice catch! We need a new character there, or we could just restore astor.

PR #29, opened 2019-04-09
Introducing organization-filtering methods for discussions

luchotc:


app/models/application_record.rb
```diff
@@ -101,8 +101,18 @@ def self.whitelist_attributes(a_hash, options={})
...
   end
 
+  def self.organic_on(*selectors)
+    selectors.each { |selector| _define_organic_on(selector) }
+  end
+
   private
 
+  def self._define_organic_on(selector)
+    define_method("#{selector}_in_organization") do |organization = Organization.current|
```
flbulgarelli: Please avoid using the `_` convention. I'd suggest inlining this method like here: https://github.com/mumuki/mumukit-core/blob/master/lib/mumukit/core/module.rb#L33

luchotc: @flbulgarelli Done! And it seems to work fine with platform 4.2

PR #30, opened 2019-04-17
Rails 5.2 & Active Storage

rodri042:


mumuki-domain.gemspec
```diff
@@ -14,7 +14,7 @@ Gem::Specification.new do |s|
 
   s.files = Dir["{app,db,lib}/**/*", "LICENSE", "Rakefile", "README.md"]
 
-  s.add_dependency "rails", "~> 5.1.6"
+  s.add_dependency "rails", "~> 5.2"
```
flbulgarelli: Always use three digits with rails versions. It does not strictly follow semver, as there may be not-so-compatible changes between minor versions.

rodri042:
> Also, I think the CI is broken since you didn't actually commit the schema after running the migrations.
Fixed!

PR #31, opened 2019-05-03
Add multifile to language's resource hash

felipecalvo:


app/models/language.rb
```diff
@@ -67,7 +67,7 @@ def to_embedded_resource_h
   def to_resource_h
     as_json(only: %i(comment_type devicon editor_css_urls editor_html_urls editor_js_urls
                     extension feedback highlight_mode layout_css_urls layout_html_urls
-                    layout_js_urls name output_content_type prompt queriable runner_url
+                    layout_js_urls multifile name output_content_type prompt queriable runner_url
```
flbulgarelli: We should add a migration, shoudn't we?

README.md
```diff
@@ -17,6 +17,7 @@
   triable
   feedback
   queriable
+  multifile
```
flbulgarelli: :+1:

app/models/language.rb
felipecalvo: I've added this migration in a recent commit.

flbulgarelli:
Related to https://github.com/mumuki/mumukit/issues/57

PR #33, opened 2019-05-09
Fix translation missing: es.submission.try_again

felipecalvo:
image

The translation wasn't working because the `submission` key was clashing with [laboratory's `submission` key](https://github.com/mumuki/mumuki-laboratory/blob/master/lib/mumuki/laboratory/locales/es.yml#L225). Therefore, I changed domain's one to `console_submission`.

I changed the locales' folder/file hierarchy as well so it's more akin to the [suggested way to do it](https://guides.rubyonrails.org/i18n.html#organization-of-locale-files).

Resolves https://github.com/mumuki/mumuki-laboratory/issues/1317


flbulgarelli:
Great! I think the narrator translations are not actually in use since the narrator gem provides their own files.

felipecalvo:
> Great! I think the narrator translations are not actually in use since the narrator gem provides their own files.

Correct, they are not in use in domain and they are the exact same as in mumukit-assistant: https://github.com/mumuki/mumukit-assistant/tree/master/lib/locales

I'm removing them now

PR #37, opened 2019-06-14
Knowing in which organization an assignment is submitted

felipecalvo:
Adds reference so we can keep track of each new assignment's organization.

Old assigmnents, in the meantime, remain with no organization

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -12,7 +12,7 @@ class Assignment < ApplicationRecord
            primary_key: :submission_id,
            dependent: :destroy
 
-  belongs_to :organization, default: -> { Organization.current }
+  belongs_to :organization
   belongs_to :submitter, class_name: 'User'
 
   validates_presence_of :exercise, :submitter
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -12,6 +12,7 @@ class Assignment < ApplicationRecord
            primary_key: :submission_id,
            dependent: :destroy
 
+  belongs_to :organization, default: -> { Organization.current }
   belongs_to :submitter, class_name: 'User'
 
   validates_presence_of :exercise, :submitter
@@ -126,7 +127,7 @@ class Assignment < ApplicationRecord
   end
 
   def to_resource_h
-    as_json(except: [:exercise_id, :submission_id, :id, :submitter_id, :solution, :created_at, :updated_at, :submission_status],
+    as_json(except: [:exercise_id, :submission_id, :organization_id, :id, :submitter_id, :solution, :created_at, :updated_at, :submission_status],
               include: {
                 guide: {
                   only: [:slug, :name],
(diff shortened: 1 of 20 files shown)


app/models/assignment.rb
```diff
@@ -126,7 +127,7 @@ def run_tests!(params)
   end
 
   def to_resource_h
-    as_json(except: [:exercise_id, :submission_id, :id, :submitter_id, :solution, :created_at, :updated_at, :submission_status],
+    as_json(except: [:exercise_id, :submission_id, :organization_id, :id, :submitter_id, :solution, :created_at, :updated_at, :submission_status],
```
flbulgarelli: I think we should expose the organization name in this resource hash

felipecalvo: The organization name is exposed on line 141, isn't it?

flbulgarelli: You are right. Dismiss my comment

app/models/assignment.rb
```diff
@@ -12,6 +12,7 @@ class Assignment < ApplicationRecord
            primary_key: :submission_id,
            dependent: :destroy
 
+  belongs_to :organization, default: -> { Organization.current }
```
flbulgarelli: Will the organization be updated if it it is resubmitted from another organization?

felipecalvo: This is changed now and it doesn't depend on default values anymore.

app/models/concerns/with_assignments.rb
```diff
@@ -18,15 +18,15 @@ def has_messages_for?(user)
     messages_for(user).present?
   end
 
-  def find_assignment_for(user)
+  def find_assignment_for(user, organization)
```
flbulgarelli: :see_no_evil: a dangling `organization` is still here

flbulgarelli: This is on purpose, right?

If that is the case, please put a comment about it, and name it `_organization`

lib/mumuki/domain/factories/organization_factory.rb
```diff
@@ -21,7 +19,20 @@
...
+    name { 'test' }
+    book { create(:book, name: 'test', slug: 'mumuki/mumuki-the-book') }
+  end
+
+  factory :another_test_organization, parent: :test_organization, traits: [:skip_unique_name_validation] do
+    book { create(:book, name: 'another-test', slug: 'mumuki/mumuki-another-book') }
+  end
+
+  trait :skip_unique_name_validation do
+    to_create { |instance| instance.save(validate: false) }
```
flbulgarelli: Didn't know `to_create` :clap:

app/models/concerns/with_assignments.rb
felipecalvo: Yes, it's all ready to filter by organization by changing
`assignments.find_by(submitter: user)`
to
`assignments.find_by(submitter: user, organization: organization)`
when [the time comes](https://github.com/mumuki/ikumi-online/issues/464).

felipecalvo: I'll change it to `_organization`, didn't know about the convention

flbulgarelli:
@felipecalvo this looks good to me. Please merge if ready.

PR #44, opened 2019-07-22
Moving Stores to Mumuki::Domain

flbulgarelli:
This PR is just for having sounder namespaces

PR #46, opened 2019-07-22
Feature custom expectations

flbulgarelli:
Merge after #44

Depends on https://github.com/mumuki/mulang/pull/247

This PR adds `custom_expectations` to guides and exercises.
It also drops `mumukit-inspection` dependency, but adds `mulang` dependency instead.


spec/models/guide_spec.rb
```diff
@@ -4,6 +4,8 @@
   let!(:extra_user) { create(:user, first_name: 'Ignatius', last_name: 'Reilly') }
   let(:guide) { create(:guide) }
 
+  it { expect(Guide.new.to_resource_h.keys).to eq Mumuki::Domain::Store::Github::GuideSchema.fields.map(&:name) }
```
flbulgarelli: @felipecalvo check this

felipecalvo: nice, so we can make sure that everything's up to date, right?

flbulgarelli: Right

app/models/guide.rb
```diff
@@ -89,9 +89,11 @@ def import_from_resource_h!(resource_h)
   # Keep this list up to date with
   # Mumuki::Domain::Store::Github::GuideSchema
   def to_resource_h
-    as_json(only: %i(beta type id_format private expectations corollary teacher_info sources learn_more authors collaborators extra settings))
+    as_json(only: %i(beta type id_format private custom_expectations corollary teacher_info sources learn_more authors collaborators extra settings))
       .symbolize_keys
       .merge(super)
+      .merge(expectations: own_expectations)
+      .merge(custom_expectations: own_custom_expectations)
```
luchotc: Is this line necessary in a guide?

app/models/exercise/problem.rb
```diff
@@ -25,12 +25,23 @@ def expectations
...
   end
 
+  def guide_custom_expectations
+    guide.custom_expectations
+  end
+
   def evaluation_criteria?
-    manual_evaluation? || own_expectations.present? || test.present?
+    manual_evaluation? ||
+      own_expectations.present? ||
```
luchotc: Maybe we can extract `own_expectations.present? || own_custom_expectations.present?` to another method

app/models/exercise.rb
```diff
@@ -117,9 +117,10 @@ def to_resource_h
     language_resource_h = language.to_embedded_resource_h if language != guide.language
     as_json(only: %i(name layout editor corollary teacher_info manual_evaluation locale
                      choices assistance_rules randomizations tag_list extra_visible goal
-                     free_form_editor_source initial_state final_state))
+                     free_form_editor_source initial_state final_state custom_expectations))
```
luchotc: Why are we using custom_expectations here if we are overriding it later? 
It also includes guide expectations which isn't necessary, right?

flbulgarelli:
Thanks for your comments @luchotc! It is ready to re-review now!

PR #50, opened 2019-09-05
adding submitted_at field for assignments

luchotc:
This was requested by @Gustrucco, so that when an assignment is updated, we can still know the real submission time.

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -44,6 +44,7 @@ class Assignment < ApplicationRecord
     transaction do
       messages.destroy_all if submission_id.present?
       update! submission_id: submission.id
+      update! submitted_at: DateTime.now
       update_submissions_count!
       update_last_submission!
     end
@@ -127,7 +128,7 @@ class Assignment < ApplicationRecord
   end
 
   def to_resource_h
-    as_json(except: [:exercise_id, :submission_id, :organization_id, :id, :submitter_id, :solution, :created_at, :updated_at, :submission_status],
+    as_json(except: [:exercise_id, :submission_id, :organization_id, :id, :submitter_id, :solution, :created_at, :updated_at, :submission_status, :submitted_at],
               include: {
                 guide: {
                   only: [:slug, :name],
--- a/db/migrate/20190905174934_add_submitted_at_to_assignments.rb
+++ b/db/migrate/20190905174934_add_submitted_at_to_assignments.rb
@@ -1,0 +1,5 @@
+class AddSubmittedAtToAssignments < ActiveRecord::Migration[5.1]
+  def change
+    add_column :assignments, :submitted_at, :datetime
+  end
+end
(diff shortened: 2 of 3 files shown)


luchotc:
I'm wondering if in [this line](https://github.com/mumuki/mumuki-domain/blob/edfdfbf0c3be0afc91274cbd0be79b6373b9772e/app/models/assignment.rb#L144) we shouldn't use submitted_at || updated_at. Do you see any downside @flbulgarelli?

app/models/assignment.rb
```diff
@@ -127,7 +128,7 @@ def run_tests!(params)
   end
 
   def to_resource_h
-    as_json(except: [:exercise_id, :submission_id, :organization_id, :id, :submitter_id, :solution, :created_at, :updated_at, :submission_status],
+    as_json(except: [:exercise_id, :submission_id, :organization_id, :id, :submitter_id, :solution, :created_at, :updated_at, :submission_status, :submitted_at],
```
flbulgarelli: Aren't we going to expose the `submitted_at` field? Why?

luchotc: Look at my comment https://github.com/mumuki/mumuki-domain/pull/50#issuecomment-528508029

app/models/assignment.rb
```diff
@@ -44,6 +44,7 @@ def persist_submission!(submission)
     transaction do
       messages.destroy_all if submission_id.present?
       update! submission_id: submission.id
+      update! submitted_at: DateTime.now
```
flbulgarelli: Perhaps we should use `DateTime.current`

See https://stackoverflow.com/questions/12748626/when-should-datetime-now-utc-vs-time-current-utc-be-used-in-rails?answertab=votes#tab-top

flbulgarelli:
> I'm wondering if in [this line](https://github.com/mumuki/mumuki-domain/blob/edfdfbf0c3be0afc91274cbd0be79b6373b9772e/app/models/assignment.rb#L144) we shouldn't use submitted_at || updated_at. Do you see any downside @flbulgarelli?

I am not sure if I understood your question. Why and where would you put that line?

app/models/assignment.rb
luchotc: Mmm, interesting, I didn't know. Nevertheless, we are using `DateTime.now` everywhere, so I'll open a separate issue for that one!

luchotc:
> > I'm wondering if in [this line](https://github.com/mumuki/mumuki-domain/blob/edfdfbf0c3be0afc91274cbd0be79b6373b9772e/app/models/assignment.rb#L144) we shouldn't use submitted_at || updated_at. Do you see any downside @flbulgarelli?
> 
> I am not sure if I understood your question. Why and where would you put that line?

That line it's already there. When notifying to 'submissions', we are doing this: 
`'created_at' => updated_at`, which makes sense, because we only care about the submission creation.
I don't think it makes sense to send Yet Another Timestamp. So i was proposing to change that, for:
`'created_at' => submitted_at || updated_at`, so assignments prior to this change aren't affected.
Then we could populate all the assignments submitted_at field with the same value as updated_at (just for the ones with nil values) and just leave:
`'created_at' => submitted_at,`, which makes more sense.

PR #55, opened 2019-10-18
Feature offline tests

flbulgarelli:
Run the following code after running migrations: 

```ruby
Exercise.where(language: Language.for_name('gobstones')).each do |exercise|
  next unless exercise.test
  
  test = YAML.load(exercise.test)
  next unless test['examples']

  examples = test['examples'].map do |example|
    { 
      initialBoard: example['initial_board'], 
      expectedBoard: example['final_board'] 
    }
  end

  exercise.update! offline_test: {examples: examples}
end
```


luchotc:
@flbulgarelli IYO, what's missing for this one to be ready to review?

flbulgarelli:
It is ready! But we should decide wether we want this feature

PR #56, opened 2019-10-22
Deprecate choice values column

julian-berbel:
:warning: This is a non backwards compatible change! :warning:
We're not using this field in any exercises though, so it doesn't really make any sense still having it there.


luchotc:
@julian-berbel If it's ok for you, let's merge it :smile:

julian-berbel:
Roger!

PR #61, opened 2019-12-17
Feature reify progress

julian-berbel:
Fixes mumuki/mumuki-laboratory#1363.


app/models/concerns/progress.rb
```diff
@@ -0,0 +1,17 @@
+module Progress
+  extend ActiveSupport::Concern
+
+  included do
+    before_save :parent, unless: :parent_id?
+    belongs_to :parent, class_name: 'Indicator', optional: true
+  end
+
+  def parent
+    super || assign_attributes(parent: parent_content&.progress_for(user, organization))
```
luchotc: Why are we using assign_attributes instead of update! ?

luchotc:
What about using the dirty flag only for content modifications and using nuntius for the submissions instead.

I imagine replacing dirty_parent! for notify_completion_changed! where we let know the parent that a childr has changed the status. There we can do something like:

```
assignment.rb

def notify_completion_changed!
   parent.notify_child_changed!
end

progress.rb

def notify_child_changed!
   # IndicatorChanged event with indicator ID
end

events

# IndicatorChanged -> find indicator && indicator.child_changed!

progress.rb

def child_changed! 
  completed_before_update? = completed?
  update! children_completed_count: ...
  parent.child_changed! if completed? != compelted_before_update?
end
```

I like this approach because it minimizes the number of dirties, making the data more "clean" (joke aside) for reports

PD: I hope this pseudo-code wasn't so pseudo

app/models/concerns/progress.rb
julian-berbel: I believe the before_save callback would cause it to go on an endless loop since update! persists the object :thinking:

luchotc: True!

luchotc:
We can have two dirty flags:
`dirty_by_content_change` and `dirty_by_submission`.

The first one won't be propagated upwards, but when it's recalculated and completion changes, the "upper level" content indicators will be updated recursively.

The second one will be propagated upwards, but when it's recalculated, it won't update "upper level" content indicators.

This will make the logic more intricate but it will be far more efficient. It also makes clear why an indicator it's dirty which is good: by submission the difference with the reality is +-1, by content changed is uncertain.

app/models/chapter.rb
```diff
@@ -26,4 +26,8 @@ def index_usage!(organization = Organization.current)
     organization.index_usage_of! topic, self
     lessons.each { |lesson| lesson.index_usage! organization }
   end
+
+  def structural_parent
```
luchotc: With this new structure we can rethink the navigation flow. I think some improvements can be made to the *Navigation mixins and to the breadcrumb logic.
We can open a separate issue for that, wdyt @julian-berbel ?

julian-berbel: Agreed!

app/models/indicator.rb
```diff
@@ -0,0 +1,78 @@
+class Indicator < ApplicationRecord
```
luchotc: I absolutely love what you've done with this class. It's far more clean and more readable.

app/models/assignment.rb
```diff
@@ -36,6 +37,20 @@ class Assignment < ApplicationRecord
     self.expectation_results = []
   end
 
+  alias_method :parent_content, :guide
+  alias_method :user, :submitter
```
flbulgarelli: This is a long awaited alias!

app/models/concerns/progress.rb
```diff
@@ -0,0 +1,17 @@
...
+    super
```
flbulgarelli: Using one-liner ifs is more intention revealing, since the or-expression value is not important: 

```ruby
assign_attributes(parent: parent_content&.progress_for(user, organization)) unless super
```

app/models/indicator.rb
```diff
@@ -0,0 +1,78 @@
...
+  private
+
+  def children
+    indicators.presence || assignments
+  end
+
+  %i(children_count children_passed_count).each do |selector|
+    define_method selector do
+      send "refresh_#{selector}!" unless self[selector]
+      rebuild!
```
luchotc: I would prefer to call rebuild! explicitly in the "entry points" of this class where we want to actually recalculate if dirty. I know that this works and it's cleaner than my solution, but I feel like we are hiding the complexity of this process. In addition, in a normal flow, rebuild! will be called twice, as in the majority of the cases we are checking both children_count and children_passed_count.  Wdyt?

app/models/concerns/progress.rb
```diff
@@ -0,0 +1,17 @@
+module Progress
```
flbulgarelli: Just wondering: why not using an abstract class here? It makes sense to think `Indicador` and `Progress` as subclasses of `Progress`.

I think that choice would be consistent with that in `Content` class: https://github.com/mumuki/mumuki-domain/blob/master/app/models/content.rb

app/models/indicator.rb
julian-berbel: you mean for example in `completion_percentage` ? yeah, that seems fair.

app/models/concerns/progress.rb
julian-berbel: > `Indicador` and `Progress` as subclasses of `Progress`

Guessing you mean `Indicator` and `Assignment` as subclasses of `Progress` ? Honestly, it didn't occur to me :stuck_out_tongue_closed_eyes: I'll go for that

flbulgarelli: :rofl:

luchotc:
Nice! Is there something else missing?
I think that for now we shouldn't notify that parent id nor the indicator per se.
We could discuss it later to see the pros and cons of it.

julian-berbel:
> Nice! Is there something else missing?
> I think that for now we shouldn't notify that parent id nor the indicator per se.
> We could discuss it later to see the pros and cons of it.

Yeah, still struggling to find an elegant way to mark indicators as dirty on content change

julian-berbel:
Still needs both unit and integration tests, but it should be pretty much done now! :sob:

app/models/concerns/with_progress.rb
```diff
@@ -10,4 +10,18 @@ def completion_percentage_for(user, organization=Organization.current)
   def dirty_progresses!
     Indicator.dirty_by_content_change! self
   end
+
+  def dirty_progress_if_children_changed
+    old_children = children.to_a
+    yield
+    Indicator.dirty_by_content_change! self if children_changed?(old_children)
+
+    self
```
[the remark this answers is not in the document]
julian-berbel: Roger!

luchotc:
Excellent :tada: :clap:. If everything works fine in the integration tests, it's more than ready to merge IMO.

app/models/concerns/with_progress.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def completion_percentage_for(user, organization=Organization.current)
+    progress_for(user, organization).completion_percentage
+  end
+
+  def dirty_progresses!
+    Indicator.dirty_by_content_change! self
+  end
+
+  def dirty_progress_if_children_changed
```
flbulgarelli: We should rename this method to `dirty_progress_if_children_changed!`

app/models/concerns/with_progress.rb
```diff
@@ -0,0 +1,27 @@
...
+    yield
+    Indicator.dirty_by_content_change! self if children_changed?(old_children)
+
+    self
+  end
+
+  private
+
+  def children_changed?(old_children)
+    Set.new(children).^(Set.new(old_children)).present?
```
flbulgarelli: It is weird to use operators this way. I think this should look like

```ruby
(Set.new(children) ^ Set.new(old_children)).present?
```

PR #62, opened 2019-12-27
Feature partial usage tree remake

julian-berbel:
Modifying usage indexation to rebuild only the necessary parts.

This should hopefully improve this process' performance a little, and we're planning on taking advantage of this on #61.

### DIFF
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -67,7 +67,7 @@ class ApplicationRecord < ActiveRecord::Base
 
   def self.aggregate_of(association)
     class_eval do
-      define_method(:rebuild!) do |children|
+      define_method("rebuild_#{association}!") do |children|
         transaction do
           self.send(association).all_except(children).destroy_all
           self.update! association => children
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -69,7 +69,7 @@ class ApplicationRecord < ActiveRecord::Base
     class_eval do
       define_method(:rebuild!) do |children|
         transaction do
-          self.send(association).all_except(children).delete_all
+          self.send(association).all_except(children).destroy_all
           self.update! association => children
           children.each &:save!
         end
--- a/app/models/concerns/guide_container.rb
+++ b/app/models/concerns/guide_container.rb
@@ -1,8 +1,10 @@
 module GuideContainer
   extend ActiveSupport::Concern
+  include WithContent
 
   included do
     validates_presence_of :guide
+    before_destroy :destroy_usages!
 
     delegate :name,
              :slug,
--- a/app/models/concerns/guide_container.rb
+++ b/app/models/concerns/guide_container.rb
@@ -4,7 +4,6 @@ module GuideContainer
 
   included do
     validates_presence_of :guide
-    before_destroy :destroy_usages!
 
     delegate :name,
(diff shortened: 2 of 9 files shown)


app/models/application_record.rb
```diff
@@ -69,7 +69,7 @@ def self.aggregate_of(association)
     class_eval do
       define_method(:rebuild!) do |children|
         transaction do
-          self.send(association).all_except(children).delete_all
+          self.send(association).all_except(children).destroy_all
```
flbulgarelli: :warning: Although using `destroy_all` is necessary for running the `destroy`  hook, this has a performance impact, since instead of running just a simple `delete from .. where ...` you will run a `select`  and then a single delete for each of the found records.

app/models/usage.rb
```diff
@@ -12,6 +12,10 @@ def self.destroy_usages_for(record)
     Usage.where(parent_item: record).destroy_all
   end
 
+  def index_children(children)
```
flbulgarelli: Perhaps it should be renamed to `index_children!`

app/models/concerns/with_usages.rb
```diff
@@ -16,6 +16,21 @@ def usage_in_organization_of_type(type, organization = Organization.current)
     item.is_a?(type) ? item : nil
   end
 
+  class_methods do
+    def content_aggregate_of(association)
+      aggregate_of association
```
flbulgarelli: Perhaps instead of introducing a new method `content_aggregate_of`, you could redefinie `agregate_of` in this mixing to add the desired behavior. That way you don't need two very similar methods, and allow using `agregate_of` without the indexig logic.

 
I think this should look like the following, thanks to mixin linearization: 

```ruby
def aggregate_of(association)
      super
     ..your code...
```

julian-berbel: yeah, that makes way more sense! :+1:

julian-berbel: actually, rethinking this, it would still be pretty obnoxious that way. I think method would end up looking something like this:
```ruby
def aggregate_of(association)
  super

  revamp :rebuild! do |_, this, children, hyper|
    old_children = this.send association
    added_children = children - old_children
    hyper.(children)
    this.usages.each { |it| it.index_children(added_children) }

    self
  end
end
```

app/models/concerns/guide_container.rb
```diff
@@ -1,8 +1,10 @@
 module GuideContainer
   extend ActiveSupport::Concern
+  include WithContent
 
   included do
     validates_presence_of :guide
+    before_destroy :destroy_usages!
```
luchotc: This one is included in WithContent

julian-berbel: oops

app/models/concerns/with_usages.rb
luchotc: Nice!

app/models/application_record.rb
luchotc: I agree with this one, but it stills improves the performance and allows us to use it in #61.
It is not that frequent to remove several children content at once, and even if we do so, it wouldn't be such a terrible workload.

luchotc:
Let's go for #61

PR #65, opened 2020-01-15
Fix orphan usages

julian-berbel:
While testing #62 I realised we were leaving some orphan usages when removing a chapter from a book, since the chapter's usages were being destroyed, but not the chapter's lessons.

This doesn't seem like an ideal solution but a quick test seems to show that this is still _way_ better than destroying and recreating the whole usage tree.

### DIFF
--- a/app/models/usage.rb
+++ b/app/models/usage.rb
@@ -7,9 +7,18 @@ class Usage < ApplicationRecord
   scope :in_organization, ->(organization = Organization.current) { where(organization_id: organization.id) }
 
   before_save :set_slug
+  before_destroy :destroy_children_usages!
+
+  def self.destroy_all_where(query)
+    where(query).destroy_all
+  end
 
   def self.destroy_usages_for(record)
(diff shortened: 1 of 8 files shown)


app/models/usage.rb
```diff
@@ -7,9 +7,18 @@ class Usage < ApplicationRecord
...
+    where(query).destroy_all
+  end
 
   def self.destroy_usages_for(record)
-    Usage.where(parent_item: record).destroy_all
+    destroy_all_where(parent_item: record)
+  end
+
+  def destroy_children_usages!
+    item.children.each { |child| Usage.destroy_all_where(item: child, organization: organization) }
```
luchotc: The performance improvement that we talked about isn't as simple as: 

`Usage.destroy_all_where(item: item.children, organization: organization)` ?

If so it would be nice to change it. If I'm missing something lets integrate it like this and leave a separate issue for that particular improvement.

PR #66, opened 2020-01-20
Chore general improvements

luchotc:
General improvements made alongside with @julian-berbel

### DIFF
--- a/app/models/concerns/navigation/siblings_navigation.rb
+++ b/app/models/concerns/navigation/siblings_navigation.rb
@@ -9,11 +9,11 @@ module SiblingsNavigation
   end
 
   def siblings
-    structural_parent.structural_children - [self]
+    structural_parent.structural_children
   end
 
   def pending_siblings_for(user, organization=Organization.current)
-    siblings.reject { |it| it.content.progress_for(user, organization).completed? }
+    siblings.reject { |it| it.progress_for(user, organization).completed? }
   end
 
   # Names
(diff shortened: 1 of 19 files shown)


julian-berbel:
photo4985594232312014950

luchotc:
Apparently `pending_exercises` was being used :stuck_out_tongue: 
We'll look into it tomorrow :smile:

app/models/concerns/with_content.rb
```diff
@@ -5,8 +5,29 @@ module WithContent
...
+
+  def progress_for(user, organization=Organization.current)
+    content.progress_for(user, organization)
+  end
+
   private
 
+  # Generally we are calling progress_for for each sibling. That method needs the
+  # content. With this includes call we're avoiding the N + 1 queries.
+  def siblings
```
flbulgarelli: Nice

app/models/concerns/navigation/siblings_navigation.rb
```diff
@@ -8,6 +8,14 @@ def restart(user)
     pending_siblings_for(user).sort_by(&:number).first
   end
 
+  def siblings
+    structural_parent.structural_children
+  end
+
+  def pending_siblings_for(user, organization=Organization.current)
```
flbulgarelli: Are se planning to change this method when indicators table is implemented?

I am wondering whether this query will become too heavy now

luchotc: It does use indicators table.
We are retrieving the associated progress ítem for each content.
In higher levels of the tree it could be very expensive in case of too many dirty indicators for that user.
In a normal flow, that's very rare though.

PR #69, opened 2020-02-19
Adding skipped to submission_status

luchotc:


lib/mumuki/domain/status/submission/skipped.rb
```diff
@@ -0,0 +1,15 @@
+module Mumuki::Domain::Status::Submission::Skipped
+  extend Mumuki::Domain::Status::Submission
+
+  def self.skipped?
+    true
+  end
+
+  def self.passed?
```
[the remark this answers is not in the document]
felipecalvo: I like this suggestion. I've been using `passed?` for lack of a `skipped` option - I hadn't taken it into account yet. I vote for `completed?` as you said.

app/models/concerns/contextualization.rb
```diff
@@ -25,7 +25,7 @@ def submission_fields_overrides
 
     delegate :visible_success_output?, to: :exercise
     delegate :output_content_type, to: :language
-    delegate :should_retry?, :to_submission_status, *Mumuki::Domain::Status::Submission.test_selectors, to: :submission_status
+    delegate :should_retry?, :to_submission_status, :completed?, *Mumuki::Domain::Status::Submission.test_selectors, to: :submission_status
```
felipecalvo: So I guess we're really using `completed?` :stuck_out_tongue:. I thought it was just a suggestion, an idea, based on your other comment.

luchotc: We are using `completed?` to make the assignment polymorphic with an indicator.
`completed?` answers if a certain progress item is, well, completed or not.

When I mentioned the thing about `passed?` I was talking about the places where we are asking it to show texts like "Congrats! You've done it". We want an skipped exercise to behave visually as a passed exercise, but it isn't technically in that status. So my idea was to have another method, like `approved?` (I realized that `completed?` is not a good one since we are using it with another purpose) and make it `passed? || skipped?`. It feels odd for an skipped exercise to answer true to `passed?`.

PR #70, opened 2020-02-21
Flow integration

felipecalvo:
:warning: **We need to tag and release flow first**

This PR adds the necessary migrations, alias and handy methods for mumukit-flow integration.

It also loses some navigation modules, which are moved to flow: https://github.com/mumuki/mumukit-flow/pull/11

Once we have a version number for flow, it should be removed from Gemfile and added to the gemspec file.


app/models/exercise.rb
```diff
@@ -36,9 +38,14 @@ class Exercise < ApplicationRecord
   validates_presence_of :submissions_count,
                         :guide, :bibliotheca_id
 
+  alias_attribute :parent, :guide
```
luchotc: We are removing the concept of both parent and children in favor of structural_* and navigable_*.
In this case, we should use structural_parent which is already defined.

app/models/guide.rb
```diff
@@ -8,11 +8,15 @@ class Guide < Content
...
           WithLanguage
 
+  include Mumukit::Flow::AdaptiveItem
+
   markdown_on :corollary, :sources, :learn_more, :teacher_info
 
   numbered :exercises
   has_many :exercises, -> { order(number: :asc) }, dependent: :destroy
 
+  alias_attribute :children, :exercises
```
luchotc: Idem for this one

app/models/guide.rb
```diff
@@ -123,6 +127,10 @@ def resettable?
     usage_in_organization.resettable?
   end
 
+  def exercise_assignments_for(user)
```
luchotc: This one seems a bit odd here, but I know it's required by flow and with the current structure there isn't a better place to implement it, so :+1:

app/models/exercise.rb
felipecalvo: So I should delete this alias (and the other one) and just use structural_parent/structural_children in flow?

app/models/guide.rb
felipecalvo: ┐(︶▽︶)┌

luchotc:
https://rubygems.org/gems/mumukit-flow :)

felipecalvo:
This is not happening soon and might have to be re-done anyway.

Issue #72, opened 2020-02-27
copy_progress_to should be transfer_progress_to


luchotc:
https://github.com/mumuki/mumuki-domain/blob/9106b5da40c785ee9fa1f93fdd700f6b7d2111c5/app/models/user.rb#L92
This method name is misleading as it is transferring the progress instead of copying it.

luchotc:
Also, it should notify the assignments. Otherwise it's prone to sync errors between apps.

PR #75, opened 2020-03-17
Renaming copy_progress_to to transfer_progress_to

luchotc:
Solves #72 and also removes resubmit! which is no longer needed


flbulgarelli:
Ok! Why isn't resubmit no longer needed? Isn´ t it necessary for... resubmissions :stuck_out_tongue: ?

PR #76, opened 2020-03-17
Chore assignments save

luchotc:
It:

- adds more tests to verify organization is properly set in assignment
- changes save! override for a callback
- changes the organization on submission

This prevents from setting an incorrect organization when updating the assignment.

The callback can be removed when all the assignments have an organization.
The line in `save_submission!` can be removed once we activate the separation of progress.

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -205,11 +205,6 @@ class Assignment < Progress
     exercise.files_for(current_content)
   end
 
-  def save!(*)
-    self.organization = Organization.current
-    super
-  end
-
   private
 
   def update_submissions_count!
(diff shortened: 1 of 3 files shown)


app/models/assignment.rb
```diff
@@ -41,6 +41,12 @@ class Assignment < Progress
   alias_method :completed?, :passed?
 
   after_save :dirty_parent_by_submission!, if: :completion_changed?
+  before_save :set_current_organization, unless: :organization
+
+  # TODO: Momentary as some assignments may not have an associated organization
+  def set_current_organization
```
flbulgarelli: Don't forget `!` in the future :smile:

luchotc: Right

PR #77, opened 2020-03-19
Add verify name logic to user

julian-berbel:


app/models/user.rb
```diff
@@ -103,6 +103,13 @@ def copy_progress_to!(another)
 
   def import_from_resource_h!(json)
     update! self.class.slice_resource_h json
+    verify_name!
```
flbulgarelli: Why are we verifying names when importing? Aren't they unrelated tasks?

julian-berbel:
There was actually a test for the opposite before our conversation :stuck_out_tongue: 
I just fixed it for the new requirement

luchotc:
Good luck it was failing then :stuck_out_tongue: 
Ready IMO

Issue #71, opened 2020-02-27
Introduce solved? concept


luchotc:
We are making `passed?` true for skipped exercises and it's not consistent. We want an skipped exercise to behave visually as a passed exercise, but it isn't technically in that status. Maybe we should introduce another concept, like `approved?` or `solved?` which represents if the exercise is, well, solved :stuck_out_tongue: ( `passed? || skipped?`) 

Note that we are using `completed?` but it is used with another purpose. It is used to make the assignment polymorphic with an indicator, as `completed?` answers if a certain progress item is, well, completed or not.

See more at #69

felipecalvo:
Here comes SOLVED-19 spreading quickly.

- mumuki-domain: https://github.com/mumuki/mumuki-domain/pull/79
- mumuki-laboratory: https://github.com/mumuki/mumuki-laboratory/pull/1384
- mumukit-core: https://github.com/mumuki/mumukit-core/pull/17
- mumuki-classroom-api: https://github.com/mumuki/mumuki-classroom-api/pull/208
- mumukit-flow: TBA

There are a couple usages of `passed?` in mumukit and mumukit-bridge, but I suspect there's no need to change anything because they're related to running stuff, and nothing will run if we're just skipping.

If there's a need to fix it there, it has to be fixed as in classroom-api - redefining the method, which is a bit dirty.

luchotc:
I think this one is `solved`, right?

PR #79, opened 2020-03-20
Skipped and passed submissions are solved

felipecalvo:
As talked about in #69 (nice) and #71

A bunch of very simple submission tests added as well

### DIFF
--- a/app/models/concerns/contextualization.rb
+++ b/app/models/concerns/contextualization.rb
@@ -25,7 +25,7 @@ module Contextualization
 
     delegate :visible_success_output?, to: :exercise
     delegate :output_content_type, to: :language
-    delegate :should_retry?, :to_submission_status, :completed?, *Mumuki::Domain::Status::Submission.test_selectors, to: :submission_status
+    delegate :should_retry?, :to_submission_status, :completed?, :solved?, *Mumuki::Domain::Status::Submission.test_selectors, to: :submission_status
     delegate :inspection_keywords, to: :exercise
   end
 
--- a/app/models/concerns/contextualization.rb
+++ b/app/models/concerns/contextualization.rb
@@ -44,11 +44,7 @@ module Contextualization
   end
 
   def results_body_hidden?
-    (passed? && !visible_success_output?) || exercise.choice? || manual_evaluation_pending?
-  end
-
-  def result_preview
-    result.truncate(100) unless passed?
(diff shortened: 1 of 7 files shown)


lib/mumuki/domain/status/submission/submission.rb
```diff
@@ -23,8 +23,6 @@ def group
     self
   end
 
-  # Tells if a new, different submission should be tried.
```
luchotc: Why? :stuck_out_tongue:

app/models/exercise/playground.rb
```diff
@@ -8,7 +8,7 @@ def setup_query_assignment!(assignment)
   end
 
   def save_query_results!(assignment)
-    unless assignment.passed?
+    unless assignment.solved?
```
luchotc: I would keep this one as just `passed?` so if a playground was skipped the student can still retry it and make it passed!.
Although we haven't thought about the interface yet, we should assume that we will let them do that.

app/models/concerns/contextualization.rb
```diff
@@ -44,11 +44,7 @@ def single_visual_result_html
   end
 
   def results_visible?
-    (visible_success_output? || !passed?) && !exercise.choices? && !manual_evaluation_pending?
-  end
-
-  def result_preview
```
luchotc: :+1: Nice catch

lib/mumuki/domain/status/submission/submission.rb
felipecalvo: I think it added nothing. Plus, the line below this one is rendered useless with tests that'll always keep it updated.

But it can come back ~unlike others~ if you think so or we can toss a virtual coin

luchotc: :+1:

app/models/exercise/playground.rb
felipecalvo: Sounds right, I'm reverting this change.

spec/models/assignment_spec.rb
```diff
@@ -231,4 +239,125 @@
     it { expect(assignment.extra).to eq "some_email\nsome_first_name\nsome_last_name\n" }
   end
 
+  describe 'submission status' do
+    let(:exercise) { create(:indexed_exercise) }
```
flbulgarelli: :clap: :clap:

felipecalvo:
I rebased because changing `results_visible?` to `results_body_hidden?` was messing with this PR.

PR #80, opened 2020-03-26
Chore hidden results for choices

luchotc:


app/models/exercise.rb
```diff
@@ -55,6 +55,10 @@ def guide_done_for?(user)
     guide.done_for?(user)
   end
 
+  def choice?
```
luchotc: We were using choices? (checks if choice field is not empty) and I think we should use this one that delegates it to the editor.

app/models/exam.rb
```diff
@@ -161,6 +161,10 @@ def limited_for?(exercise)
     max_attempts_for(exercise).present?
   end
 
+  def results_hidden_for?(exercise)
```
luchotc: For now only choices can be limited in an exam

app/models/concerns/contextualization.rb
```diff
@@ -43,14 +43,27 @@ def single_visual_result_html
...
-  def results_visible?
-    (visible_success_output? || !passed?) && !exercise.choices? && !manual_evaluation_pending?
+  # It tells whether the results body should be hidden or not.
+  # The results body shouldn't be displayed when:
+  # * The contextualization passes and it shouldn't output anything on success.
+  # * It's a choice exercise.
+  # * It hasn't been evaluated manually yet.
+  def results_body_hidden?
+    (passed? && !visible_success_output?) || exercise.choice? || manual_evaluation_pending?
   end
```
flbulgarelli: Does this mean that choices result will never be displayed again, right?

flbulgarelli: What will happen with choice exercises with assistance rules?

app/models/exercise.rb
flbulgarelli: OK

app/models/concerns/contextualization.rb
luchotc: This one it's just a rename. The logic is the same as it was before, I've just change the conditions to make it more intention_revealing.
This doesn't affect the assistance_box that is always _tried_ to be displayed when the assignment doesn't passes

app/models/exercise.rb
```diff
@@ -198,7 +198,7 @@ def try_submit_solution!(user, solution={})
 
   # An exercises with hidden results cannot be limited
```
felipecalvo: Here's another one,  'exerciseS'

app/models/concerns/contextualization.rb
felipecalvo: I know I may be raising the no-comment flag a bit too much lately, but still: is there anything on line 52 that is not clear enough as to warrant a five-line comment?

The domain logic behind whether we should hide the results body or not may be complex, but the resulting code isn't.

Still, if you think it's not clear enough, maybe `(passed? && !visible_success_output?)` could be extracted into `silent_output_if_passed?`, or something of the sort.

felipecalvo: I'm not against comments just for the sake of it, of course - for example I think the one you added at https://github.com/mumuki/mumuki-domain/pull/80/commits/8e90a8b7d9eb1f243520e0b5c5584a938982196d#diff-a846517fde5cc3ecf4655c73a3a7a09bR199 is much more useful because the underlying justification for adding `!results_hidden?` can't be derived from reading the code.

luchotc: TBH I've written the comment first as the code wasn't obvious to me and then refactor it so I didn't recheck if the comment was useful or not.
It was, but just to rewrite the method, so I agree on removing it :smile:

PR #84, opened 2020-05-12
Feature user avatars

felipecalvo:


app/models/user.rb
```diff
@@ -140,14 +140,22 @@ def currently_in_exam?
...
     self.uid ||= email
   end
 
   def init
-    self.image_url ||= "user_shape.png"
+    self.avatar = random_avatar unless profile_picture.present?
+  end
+
+  def random_avatar
+    Avatar.order('RANDOM()').first
```
felipecalvo, line 158: This doesn't cause performance issues with few (< thousands) rows. I don't think we'll have more than a couple dozens, so it should be okay.

app/models/user.rb
```diff
@@ -138,14 +140,18 @@ def currently_in_exam?
     exams.any? { |e| e.in_progress_for? self }
   end
 
+  def profile_picture
+    (self.avatar || self)&.image_url
```
luchotc, line 144: I'll prefer `avatar&.image_url || image_url`
I find that notation quite cumbersome

lib/mumuki/domain/helpers/user.rb
```diff
@@ -63,6 +63,10 @@ def full_name
 
   alias_method :name, :full_name
 
+  def random_avatar
+    Avatar.order('RANDOM()').first
```
luchotc, line 67: This a curious problem that I've never dealt with.
I'm sure you've read the docs but just to make it explicit here:

This method is the fastest but it has the downside of not being portable nor between different DBs nor between Rails versions (doesn't work in Rails 6)

`Avatar.offset(rand(Avatar.count)).first` is the alternative but it's a bit more verbose and inefficient.

Thinking a bit ahead, we'll probably change this one throughout the development of the feature thinking that _maybe_ not all avatars are initially available.
I'd leave it this way with a comment for now. 

Having said that, I'd move this method to Avatar and call `Avatar.random` here.

db/migrate/20200508195858_add_avatar_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddAvatarToUsers < ActiveRecord::Migration[5.1]
+  def change
+    add_reference :users, :avatar, index: true
```
luchotc, line 3: I'm not sure an index is necessary here as I don't think we'll filter users based on their avatar, right?

lib/mumuki/domain/helpers/user.rb
flbulgarelli: Or even better, `Avatar.sample`, which is consistent with collections naming: `[1, 2, 3, 4].sample` returns an element in a random position

luchotc: Definitely `sample` is more consistent!

app/models/user.rb
felipecalvo: Will change

db/migrate/20200508195858_add_avatar_to_users.rb
felipecalvo: I indexed it as most (if not all) foreign keys were indexed too, but your observation is right. I'll rewrite this migration so it's cleaner.

felipecalvo:
Done:

No index at https://github.com/mumuki/mumuki-domain/pull/84/commits/bf47661ad85e6880f96647255ed66b5e895669bc

Sample avatar and simpler profile_picture at https://github.com/mumuki/mumuki-domain/pull/84/commits/65511fd9ff6420e1c778a9befb1f133456f24d65

luchotc:
Ready to merge IMO :smile:

PR #85, opened 2020-05-12
Adding archive_from to organization settings

luchotc:


flbulgarelli:
:clap: :clap:

lib/mumuki/domain/organization/settings.rb
```diff
@@ -24,4 +28,12 @@ def login_methods
   def forum_discussions_minimal_role
     (@forum_discussions_minimal_role || 'student').to_sym
   end
+
+  def archived?
+    archiving_enabled? && archive_from.to_datetime < DateTime.now
+  end
+
```
flbulgarelli, line 35: Does it make sense organization to be archived and disabled?

I think we are missing something like a global state of organization.

luchotc: We'll introduce in a new one the semantics related to `in_preparation?` (not defined yet :stuck_out_tongue: ).
Maybe there it'll make sense to reify them in 3 states.
Pre -> 403 and "The course will start X"
200
Post -> 410 and "The course has finished".

Do you another possible state?

flbulgarelli: Yup, that sound goods. I'd introduce an `organization.status` with three possible values.

luchotc:
@flbulgarelli I could use a re-review here! I've decided not to reify the status in organization as its behavior it's very limited yet. I think that it would be needed when archived state is introduced but we are OK with what we have meanwhile.

PR #87, opened 2020-05-14
Feature affable methods

flbulgarelli:
### :dart: PR Goal

This PR introduces the new general concept of `affable`, and provides implementations for strings and `nil`. 

Affable is modeled thinking of the `humanized` concept of rails, but... with a different name (:stuck_out_tongue: ) and semantics: it is abut converting things with markdown into santized, html-compatible things. 

`affable` concept thus reifies some previously methods known as `*_html` or `humanized_*`, and introduces the concept of a 'htmelization' pipeline: 

1. translate
2. `markdownified` convert into markdown
3.  `sanitized`: convert the object into one with identical structure, but sanitized content
4. `affable`: hide low level details in the structure, changing it if necessary
5. convert into html 

The two first steps are domain-level operations, but the latter is UI-specific, so it is outside this PR. 

Those methods may be implemented in any kind of objects or attributes. In strings they are implemented so that `affable` is just `markdownified` + `sanitized`. Implementation is nil-safe. 

## :back: Backward compatibility

This PR is fully backward compatible - aside of some deprecation warnings. It should be introduced in a minor release

## :memo:  Additional notes

I made markdown conversion names a bit more consistent - still, the `markdown_on` produces methods like `*_html` instead of `markdownified_*`, that may be refactored in the future - and added missing tests. Also, this is the first code in domain that explicitly uses `summary` field from bridge. 

### :eyes: Relations

 * Related to https://github.com/mumuki/mumukit-assistant/pull/7 - because I had problems testing this feature
* Related to https://github.com/mumuki/mumuki-laboratory/pull/1400 - because I am using this features in it

### DIFF
--- a/app/models/concerns/contextualization.rb
+++ b/app/models/concerns/contextualization.rb
@@ -98,6 +98,7 @@ module Contextualization
   end
 
   def humanized_expectation_results
+    warn "Don't use humanized_expectation_results. Use affable_expectation_results, which also handles markdown an sanitization"
     visible_expectation_results.map do |it|
       {
         result: it[:result],
--- a/app/models/concerns/contextualization.rb
+++ b/app/models/concerns/contextualization.rb
@@ -107,7 +107,6 @@ module Contextualization
     end
   end
 
-
   ####################
   ## Affable results
   ####################
@@ -127,12 +126,12 @@ module Contextualization
 
   def affable_test_results
     test_results.map do |it|
-      {
-        title: test_result[:title].affable,
-        result: test_result[:result].sanitized,
-        status: test_result[:status],
-        summary: test_result.dig(:summary, :message).affable
-      }
+      { summary: it.dig(:summary, :message).affable }
+        .compact
+        .merge(
+          title: it[:title].affable,
+          result: it[:result].sanitized,
+          status: it[:status])
     end
   end
 end
--- a/app/models/concerns/contextualization.rb
+++ b/app/models/concerns/contextualization.rb
@@ -105,4 +105,33 @@ module Contextualization
       }
     end
   end
+
+
+  ####################
+  ## Affable results
+  ####################
+
+  def affable_expectation_results
+    visible_expectation_results.map do |it|
+      {
+        result: it[:result],
+        explanation: Mulang::Expectation.parse(it).translate(inspection_keywords).affable
+      }
+    end
+  end
+
+  def affable_tips
+    tips.map(&:affable)
+  end
+
+  def affable_test_results
+    test_results.map do |it|
+      {
+        title: test_result[:title].affable,
+        result: test_result[:result].sanitized,
+        status: test_result[:status],
+        summary: test_result.dig(:summary, :message).affable
+      }
+    end
+  end
 end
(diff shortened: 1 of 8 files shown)


lib/mumuki/domain/extensions/hash.rb
```diff
@@ -1,5 +1,14 @@
 class Hash
-  def markdownify!(*keys)
-    keys.each { |it| self[it] = Mumukit::ContentType::Markdown.to_html(self[it]) }
+  def markdownify!(*keys, **options)
+    warn "Don't use markdownify. Use markdownified! instead"
+    markdownified! *keys, **options
+  end
+
+  def markdownified!(*keys, **options)
+    keys.each { |it| self[it] = self[it].markdownified }
```
julian-berbel, line 8: is it ok for options to be ignored here? :monocle_face:

flbulgarelli: nope!!! My bad

flbulgarelli: Solved!

app/models/concerns/contextualization.rb
```diff
@@ -98,6 +98,7 @@ def visible_expectation_results
   end
 
   def humanized_expectation_results
+    warn "Don't use humanized_expectation_results. Use affable_expectation_results, which also handles markdown an sanitization"
```
felipecalvo, line 101: markdown _and_ sanitization

lib/mumuki/domain/extensions/string.rb
```diff
@@ -38,15 +38,24 @@ def file_extension
 #  * markdownified: interpretes markdown in message and generates HTML
 #  * sanitized: sanitizes results HTML
 #  * affable: changes structure to hide low level details
+#
+# Other classes may polymorphically implement they own
```
felipecalvo, line 42: their :smile:

lib/mumuki/domain/extensions/string.rb
```diff
@@ -38,15 +38,24 @@ def file_extension
...
+# markdownified, sanitized and affable methods with similar semantics
+# to extend this pipeline to non-strings
 class String
+
+  # Creates a humman representation - but not necessary UI - representation
+  # of this strings by interpreting its markdown as a one-liner and sanitizing it
```
felipecalvo, line 48: this _string_ :100:

flbulgarelli:
> So clean 
> 
> I'm just not so sure about the name affable - friendly sounds more natural to me - but we have `friendly` all over the codebase so I guess it's okay.

Yeah that is exactly the issue. Both `friendly` and `humanized` are already taken. I tried to find a synonym of a _kind, easy-goging, gentle_ thing. Affable is actually something that has sense to me even in Spanish - _afable_ means exactly the same - but it is not a so-common word. 

Actually my other candidates where `kind` - but it also has a type-semantics - and `gentle` but I think it resembles more to _good natured_ than _easy to deal with_

luchotc:
I agree that `affable` wouldn't be my first choice but we don't have many other options.
What about `pretty`? It's quite a known semantic but it's more related to other aspects of the string display. Nevertheless, it could be an option...

flbulgarelli:
Great! I will fix comments and merge. 

I didn't think in `pretty`!  I think It could have been a good candidate, but I feel it quite related to pretty-printing code and pretty-printing objects for the console.

PR #90, opened 2020-05-28
Feature user delete

flbulgarelli:
# :dart: Goal

This PR introduces the concept of :ghost:  disabled users: users that can not navigate the site, and whose public profile is replaced by a generic "ghost" identity. 

It also introduced the concept of a _buried_profile_ which represents such phantasmagorical identity, in order to be able to decouple those changes of the selection of those names.  \

# :soon: Future work

I could have introduced also a generic i18n key, so that that identity could also be translated, but that would introduce a lot of rendering issues in laboratory, classroom and maybe bibliotheca. So the identity will be universal and not localizable by now. 

# :eyes: Relations

* Related to https://github.com/mumuki/mumuki-domain/issues/86. This PR implements some of its ideas
  * _user should not be able to login again, even if the login provider permits it_: it allows login, but then kicks out user
  * _user could be displayed with a standard "deleted account" username and image_: profile image override is not covered here, but name, gender and birthdate are
  *  _No new notifications should be sent by email_: this PR buries them

# :memo: Additional notes

The concept of _disabling_ is introduced in a generic way, to allow future uses of disabled resources  - for example, content items. In order to do that, a `bury!` hook is introduced, to perform disabling editions to a resource. 

This PR adds a new feature, so it should go to a **minor**, but introduces non-breaking changes. Even if you don't configure the `buried_profile`, everything will work.


spec/spec_helper.rb
```diff
@@ -52,3 +52,5 @@ def reindex_current_organization!
 end
 
 SimpleCov.start
+
+User.configure_buried_profile! first_name: 'shibi', last_name: '', email: 'shibi@mumuki.org'
```
flbulgarelli, line 56: :ghost:  死び , the dead cousin of :yellow_heart: きび, :blue_square: めび and :red_circle:  よび

app/models/concerns/disabling.rb
```diff
@@ -0,0 +1,37 @@
+# The disposable module is a soft-delete helper that:
+#
+#  * adds `disable!` method that set a `disabled_at` attribute anden then _buries_ the object
```
luchotc, line 3: typo :train2:

spec/models/user_spec.rb
```diff
@@ -263,4 +263,33 @@
     it { expect(student_not_in_exam.currently_in_exam?).to be false }
     it { expect(student_in_exam.currently_in_exam?).to be true }
   end
+
+
+  describe 'disabling' do
+    let(:user) { create(:user, first_name: 'John', last_name: 'Doe') }
+
+    shared_context "soft delete" do
```
luchotc, line 271: Really useful. We should use it more!

spec/models/user_spec.rb
```diff
@@ -263,4 +263,33 @@
...
+      it_behaves_like "soft delete"
+    end
+
+    describe '#destroy!' do
+      before { user.destroy! }
+      it_behaves_like "soft delete"
+    end
+
+    describe '#destroy' do
+      before { user.destroy! }
```
luchotc, line 291: `user.destroy` ?

app/models/concerns/disabling.rb
```diff
@@ -0,0 +1,37 @@
...
+    !disabled?
+  end
+
+  # override to perform additional
+  # post-disable actions
+  def bury!
+  end
+
+  def ensure_enabled!
+    raise Mumuki::Domain::ForbiddenError if disabled?
```
luchotc, line 31: In some contexts I think a NotFound is more accurate but I'm assuming that we are using this to represent that it's beyond user's scope, so :+1:

app/models/user.rb
```diff
@@ -144,6 +145,11 @@ def profile_picture
     avatar&.image_url || image_url
   end
 
+  def bury!
```
luchotc, line 148: :100:

app/models/concerns/disabling.rb
flbulgarelli: I actually would like to use a new custom exception, `Mumuki::Domain::DisabledResourceError`, that would generate a 503 error in laboratory. Does it make sense?

luchotc: Totally!

PR #91, opened 2020-05-28
Chore avatar in resource hash

felipecalvo:
I changed the user resource hash _a little too much_ in https://github.com/mumuki/mumuki-domain/pull/84, which would inadvertently break/cause inconsistencies between labo/academia and classroom. (inadvertently to me :stuck_out_tongue_closed_eyes:)

Additionally, it temporarily rollbacks setting avatars to new users so we can safely deploy at this intermediate stage in which the avatars are still to be decided on, drawn and added.

Closes https://github.com/mumuki/ikumi-online/issues/625

Just merge if it's okay

### DIFF
--- a/spec/lib/user_helpers_spec.rb
+++ b/spec/lib/user_helpers_spec.rb
@@ -45,7 +45,7 @@ describe Mumuki::Domain::Helpers::User do
       it { expect(user.to_resource_h)
                .to json_eq(
                        uid: json[:uid],
-                       profile_picture: json[:image_url],
+                       image_url: json[:image_url],
                        email: json[:email],
                        first_name: json[:first_name],
                        last_name: json[:last_name],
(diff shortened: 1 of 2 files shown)


spec/lib/user_helpers_spec.rb
```diff
@@ -45,7 +45,7 @@
       it { expect(user.to_resource_h)
                .to json_eq(
                        uid: json[:uid],
-                       profile_picture: json[:image_url],
+                       image_url: json[:image_url],
```
flbulgarelli, line 48: Could we just keep both? Having avatars would be nice to https://github.com/mumuki/mumuki-domain/pull/90

luchotc: Take into account that image_url is defined as `profile_picture` in resource_h which does use avatars.
For now no avatar is set, but that's because we wanted to avoid having to create new avatars in the db right before this one is _integrated_.

PR #95, opened 2020-06-08
Feature progressive display

flbulgarelli:
# :dart: Goal 

This the first step in the stair to having customized access rules. This PR adds simple support for only displaying contents that go next after those that have been completed. 

# :memo: More details

This PR is related to https://github.com/mumuki/mumuki-domain/compare/feature-access-rules, which generalizes this concepts, and extends it to other kind of rules, courses and allows better customization of the progressive display. 

I also performed some refactors not included on that branch, and introduced the concept of `Area`, that is a common interface for `Organization` and `Course` - and maybe feature organizational divisions in the future. 

# :soon: Future work

* Allow this behaviour to also be customizable by course
* Extend this behaviour to lessons, not just chapters
* Introduce distinction between hidden and disable contents. For now this will be a ui-specific task

This PR is mostly backward compatible, but renames the `WithContent` to `Container` , which is a more semantic terms that is already used in the platform. Also, this is a new feature, so it should require its own minor version.  

It also refactors the `make_*_of!` and `*_of?` methods in a way that _should_ be backward compatible, except for flaky tests that use `structs`  instead of actual domain objects.


lib/mumuki/domain/helpers/user.rb
```diff
@@ -49,9 +49,9 @@ def discusser_of?(organization)
   (Mumukit::Auth::Roles::ROLES - [:owner]).each do |role|
 
     # Assignes the #{role} role to this user
-    # for the given slug
-    define_method "make_#{role}_of!" do |slug|
-      add_permission! role, slug
+    # for the given `grant_like`
+    define_method "make_#{role}_of!" do |grant_like|
+      add_permission! role, grant_like.to_mumukit_grant
```
flbulgarelli, line 54: This change is not actually necessary, but I added `to_mumukit_grant` to make the conversion more explicit

spec/models/book_spec.rb
```diff
@@ -200,4 +200,121 @@
       it { expect(Chapter.count).to eq 1 }
     end
   end
+
+  describe 'enabled chapters' do
```
luchotc, line 204: I don't know if it's intentional, but a lookahead of 0 behaves like fully display. I think it's the expected behavior but we should either add a test for it or validate that only positive values are given.

flbulgarelli: Nice catch.

Yeah, it is an expected behaviour, because that would mean otherwise that you can't see even the content you are working in. 

I'll add a test and a validation and merge, does it make sense?

luchotc: Yesss :smile:

PR #97, opened 2020-06-16
Feature classroom model unification

fedescarpa:


luchotc:
@flbulgarelli Would like to hear your thoughts about this one too

app/models/course.rb
```diff
@@ -14,24 +14,22 @@ class Course < ApplicationRecord
...
   def import_from_resource_h!(resource_h)
     update! self.class.slice_resource_h(resource_h)
   end
 
   def slug=(slug)
-    s = Mumukit::Auth::Slug.parse(slug)
-
-    self[:slug] = slug
-    self[:code] = s.course
+    s = slug.to_mumukit_slug
```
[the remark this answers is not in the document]
fedescarpa: This change should be implemented into `use-classroom-as-gem` branch

app/models/exam.rb
```diff
@@ -116,10 +125,41 @@ def reindex_usages!
...
+
+  def results_hidden_for?(exercise)
+    exercise.choice? && results_hidden_for_choices?
+  end
+
+  def resettable?
+    false
+  end
+
+  def set_classroom_id
```
julian-berbel, line 144: missing `!` convention :slightly_smiling_face:

app/models/exam.rb
```diff
@@ -116,10 +125,41 @@ def reindex_usages!
...
+
+  def passing_criterion
+    @passing_criterion ||= Exam::PassingCriterion.parse(passing_criterion_type, passing_criterion_value)
+  end
+
+  def ensure_valid_passing_criterion!
+    passing_criterion.ensure_valid!
+  end
+
+  def set_default_criterion_type
```
julian-berbel, line 156: same here

app/models/invitation.rb
```diff
@@ -2,12 +2,18 @@ class Invitation < ApplicationRecord
...
   belongs_to :course
+
+  validate :ensure_not_expired, on: :create
   validates_uniqueness_of :code
 
   defaults do
     self.code ||= self.class.generate_code
   end
 
+  def ensure_not_expired
```
julian-berbel, line 13: same here?

luchotc: In general we are using `!` when the `ensure_` explicitly raises an exception

app/models/course.rb
```diff
@@ -41,9 +39,9 @@ def closed?
     current_invitation.blank? || current_invitation.expired?
   end
 
-  def generate_invitation!(expiration_date)
-    invitation = invitations.build expiration_date: expiration_date, course: self
-    invitation.save_and_notify!
+  def create_invitation_for(expiration_date)
```
flbulgarelli, line 42: I think the `for` preposition sounds a little awkward here, since the given date is not _the recipient_ of the invitation.  Why don't `create_invitation`, `generate_invitation`, or just `invite!`?

app/models/exam.rb
```diff
@@ -1,19 +1,28 @@
 class Exam < ApplicationRecord
+
   include GuideContainer
   include FriendlyName
-
-  validates_presence_of :start_time, :end_time
+  include TerminalNavigation
```
flbulgarelli, line 5: Nice

app/models/exam.rb
```diff
@@ -1,19 +1,28 @@
...
 
   belongs_to :organization
+  belongs_to :course, optional: true
 
   has_many :authorizations, class_name: 'ExamAuthorization', dependent: :destroy
   has_many :users, through: :authorizations
 
+  enum passing_criterion_type: [:none, :percentage, :passed_exercises], _prefix: :passing_criterion
```
flbulgarelli, line 13: Why not just calling it `passing_criterion`? Is there any difference between the criterion and the criterion type?

app/models/exam.rb
```diff
@@ -1,19 +1,28 @@
...
+
+  validates_presence_of :start_time, :end_time
+  validates_numericality_of :max_problem_submissions, :max_choice_submissions, greater_than_or_equal_to: 1, allow_nil: true
+
+  before_save :set_default_criterion_type!
+  before_save :ensure_valid_passing_criterion!
+
+  before_create :set_classroom_id!
```
flbulgarelli, line 21: :thinking: 

I am not quite sure about the semantics of a _classroom id_ in domain. Why not calling it `code`?

app/models/exam.rb
```diff
@@ -141,8 +181,13 @@ def self.upsert_students!(json)
...
     exam[:guide_id] = Guide.locate!(exam[:slug]).id
     exam[:organization_id] = Organization.current.id
+    exam[:course_id] = Course.locate!(exam[:course].to_s).id
     exam[:users] = User.where(uid: exam[:uids])
-    [:start_time, :end_time].each { |param| exam[param] = exam[param].to_time }
+    exam[:start_time] = exam[:start_time].to_time
+    exam[:end_time] = exam[:end_time].to_time
+    exam[:classroom_id] = exam[:eid] if exam[:eid].present?
+    exam[:passing_criterion_type] = exam.dig(:passing_criterion, :type)
+    exam[:passing_criterion_value] = exam.dig(:passing_criterion, :value)
```
flbulgarelli, line 190: Perhaps we could use plain setters here: 

```ruby
exam.guide = Guide.locate!(exam.slug)
exam.organization = Organization.current
exam.course = Course.locate!(exam.course.to_s)
# and so on...
```

app/models/exam/passing_criterion.rb
```diff
@@ -0,0 +1,53 @@
+class Exam::PassingCriterion
+
+  attr_reader :value
+
+  def initialize(value)
+    @value = value
+  end
+
+  def type
+    self.class.name.demodulize.underscore
```
flbulgarelli, line 10: Here you could use mumukit-core `as_variable_name`

app/models/exam.rb
luchotc: Yep, I agree. I would leave it for the next PR though, as it actually unifies everything and makes this change more straight forward

app/models/exam.rb
luchotc: Mmm I agree with you but classroom_id it's already a field in domain :stuck_out_tongue: 
We should change it in the next PR, that's true

app/models/exam/passing_criterion.rb
luchotc: Mmm I think it wouldn't work as it doesn't demodulize it :confounded:

PR #99, opened 2020-06-24
Feature experience points

felipecalvo:
# :dart: Goal 

This PR awards experience points to users when an exercise is solved successfully.

# :memo: Details

Requires a new entity, called `ExperiencePoints`, which keeps track of how many exp points an user has on an organization.

100 points are awarded for passed exercises, while 50 are awarded for those passed with warnings. If an exercise passed with warnings is fixed, 50 more points are awarded - capping at 100 no matter how many tries it took or how they looked like.

This what-was-your-personal-best? on an exercise required adding a new field to assignments, called `top_submission_status`. It stores the best result ever on that assignment, whether it was passed with warnings, passed or skipped. A (positive) side effect of this is we won't award infinite points if an exercise is passed and then failed and then passed an indeterminate amount of times - it's just a matter of checking what the best result was.

# :question: Missing

~I'm writing tests.~ Done

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -40,6 +40,7 @@ class Assignment < Progress
   alias_method :parent_content, :guide
   alias_method :user, :submitter
 
+  before_save :award_experience_points!, :update_top_submission!, if: :submission_status_changed?
   after_save :dirty_parent_by_submission!, if: :completion_changed?
   before_validation :set_current_organization!, unless: :organization
 
@@ -227,7 +228,7 @@ class Assignment < Progress
   end
 
   def update_top_submission!
-    self.update! top_submission_status: submission_status unless submission_status.improved_by?(top_submission_status)
+    self.top_submission_status = submission_status unless submission_status.improved_by?(top_submission_status)
   end
 
   private
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -160,7 +160,10 @@ class Assignment < Progress
   end
 
   def to_resource_h
-    as_json(except: %i(exercise_id submission_id organization_id id submitter_id solution created_at updated_at submission_status submitted_at parent_id),
+    excluded_fields = %i(created_at exercise_id id organization_id parent_id solution submission_id
+                         submission_status submitted_at submitter_id top_submission_status updated_at)
+
+    as_json(except: excluded_fields,
               include: {
                 guide: {
                   only: [:slug, :name],
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -226,6 +226,10 @@ class Assignment < Progress
     exercise.files_for(current_content)
   end
 
+  def update_top_submission!
+    self.update! top_submission_status: submission_status unless submission_status.improved_by?(top_submission_status)
+  end
+
   private
 
   def update_submissions_count!
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -1,6 +1,7 @@
 class Assignment < Progress
   include Contextualization
   include WithMessages
+  include Gamified
 
   markdown_on :extra_preview
 
--- a/lib/mumuki/domain/status/submission/passed.rb
+++ b/lib/mumuki/domain/status/submission/passed.rb
@@ -8,4 +8,8 @@ module Mumuki::Domain::Status::Submission::Passed
   def self.iconize
     {class: :success, type: 'check-circle'}
   end
+
+  def self.exp_given
+    100
+  end
 end
--- a/lib/mumuki/domain/submission/base.rb
+++ b/lib/mumuki/domain/submission/base.rb
@@ -48,8 +48,6 @@ class Mumuki::Domain::Submission::Base
   def save_results!(results, assignment)
     assignment.assign_attributes results
     assignment.increment_attempts!
-    assignment.award_experience_points!
-    assignment.update_top_submission!
     assignment.save! results
   end
 
--- a/lib/mumuki/domain/submission/base.rb
+++ b/lib/mumuki/domain/submission/base.rb
@@ -48,6 +48,7 @@ class Mumuki::Domain::Submission::Base
   def save_results!(results, assignment)
     assignment.assign_attributes results
     assignment.increment_attempts!
+    assignment.award_experience_points!
     assignment.save! results
   end
 
--- a/lib/mumuki/domain/submission/base.rb
+++ b/lib/mumuki/domain/submission/base.rb
@@ -49,6 +49,7 @@ class Mumuki::Domain::Submission::Base
     assignment.assign_attributes results
     assignment.increment_attempts!
     assignment.award_experience_points!
+    assignment.update_top_submission!
     assignment.save! results
   end
 
--- a/app/models/concerns/contextualization.rb
+++ b/app/models/concerns/contextualization.rb
@@ -16,6 +16,7 @@ module Contextualization
   end
 
   included do
+    serialize :top_submission_status, Mumuki::Domain::Status::Submission
     serialize :submission_status, Mumuki::Domain::Status::Submission
     validates_presence_of :submission_status
 
--- a/app/models/concerns/gamified.rb
+++ b/app/models/concerns/gamified.rb
@@ -1,0 +1,11 @@
+module Gamified
+  def award_experience_points!
+    stats = user_stats_for(submitter, organization)
+    stats.exp += submission_status.exp_given
+    stats.save!
+  end
+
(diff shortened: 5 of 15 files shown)


spec/models/with_gamification.rb
```diff
@@ -0,0 +1,90 @@
+require 'spec_helper'
+
+describe WithGamification, organization_workspace: :test do
+
+  def award_experience_points!(assignment, status)
+    assignment.status = status
+    assignment.award_experience_points!
+    assignment.update_top_submission!
+  end
```
felipecalvo, line 9: This is not ideal because it depends on order: points must be awarded before updating the top submission, hence this helper method.

One way of avoiding this would be updating the top submission _inside_ `award_experience_points!` but it doesn't feel right to me.

Right now it works like this https://github.com/mumuki/mumuki-domain/blob/78209fb65cc4ffebb137987956ab83b20804b47f/lib/mumuki/domain/submission/base.rb#L48-L54

but the order of execution is not tested, as points awarding and submission updating are tested individually.

One possible option to enforce this order would be also testing experience points on this set of specs

https://github.com/mumuki/mumuki-domain/blob/c31a9c3a65c9a3017f026ec2fe086c97da6bb55d/spec/models/assignment_spec.rb#L369

which already mimic submitting a solution (therefore going through `save_results!`), but I thought it'd get a little messy - too many things at once. I'm hearing opinions.

lib/mumuki/domain/status.rb
```diff
@@ -1,6 +1,8 @@
 module Mumuki::Domain::Status
   extend ActiveSupport::Concern
 
+  include Comparable
```
luchotc, line 4: I like the idea of having ..::Skipped > ..::Failed but 
a) I'm not sure if it's semantically meaningful
b) It could lead to some unexpected behavior:
The status has a custom implementation of `==` but if it didn't this would cause `Skipped == Passed` as it would compare them based on the exp points.

I think your first implementation is a bit more declarative. Maybe `tops?` or `improves?` are good alternative as well.

luchotc: Just to be clear: I love the fact that the comparison is done based on the exp_points. I would implement the method the same way, I'd just use something custom instead of `>`

db/migrate/20200616160640_create_experience_points.rb
```diff
@@ -0,0 +1,10 @@
+class CreateExperiencePoints < ActiveRecord::Migration[5.1]
```
luchotc, line 1: I find calling `experience_points.exp` not ideal but I'm not sure what is the ideal alternative.

`experience.points` maybe be better, but calling the entity just Experience doesn't seem ideal either.
`experience_points.total` may be an alternative :thinking: 

We could think a bit ahead and use this entity to store the `streaks` as well, and make it more than it is now. With that in mind maybe we could come up with better names, like ¿stats?.

spec/models/with_gamification.rb
luchotc: I would prefer to use the second option. 
Adding that helper method may cause some scenarios to pass where in reality they don't.

You could test this in the other spec and remove this one if you think it's less messy :smile:

felipecalvo: I'm not 100% sure where I like it the most. I'll merge them in one single spec file, I'll think which one.

app/models/concerns/with_gamification.rb
```diff
@@ -0,0 +1,19 @@
+module WithGamification
+  def award_experience_points!
+    points = net_experience
+
+    if points > 0
+      xp = experience_points_for(submitter, organization)
```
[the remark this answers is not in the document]
felipecalvo: Didn't know `increment!` was a thing.

lib/mumuki/domain/status.rb
felipecalvo: Alright. I first did `lesser_than?` as you saw but I hated the name. I'll revert this commit and change the original implementation to `improves?`.

lib/mumuki/domain/submission/base.rb
```diff
@@ -48,6 +48,8 @@ def save_submission!(assignment)
   def save_results!(results, assignment)
     assignment.assign_attributes results
     assignment.increment_attempts!
+    assignment.award_experience_points!
```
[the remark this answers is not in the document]
felipecalvo: Oof, forgot about `skipped`. Good catch. I think moving both lines to a `before_save` check should work.

db/migrate/20200616160640_create_experience_points.rb
felipecalvo: I was thinking about the name as well because `experience_points.exp` is not very good.

We have a couple other entities which represent a many-to-many users and organizations relationship, but they were created to support a third entity: an exercise in `assignments`, a parent content in `indicators`, a discussion in... `discussions`.

I think we should strive to come up with a new domain name for this, which conveys "this user in this organization". I'm not a fan of `stats` because it's such a broad term.

A word that represents `one who follows a path` would be my most preferred choice (we have mumuki.io/paths, don't we?) but I'm not high enough to choose `Traveller`, `Explorer` or `Voyager`. We're not that dangerous either.

image

app/models/assignment.rb
```diff
@@ -1,6 +1,7 @@
 class Assignment < Progress
   include Contextualization
   include WithMessages
+  include WithGamification
```
flbulgarelli, line 4: Why not just `Gamified`?  I think it makes more sense in English, doesn't it?

app/models/concerns/with_gamification.rb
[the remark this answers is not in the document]
flbulgarelli: It is a nice method, but just be aware that `increment! ` does not update `updated_at`. I had troubles with it in the past

lib/mumuki/domain/status/submission/passed.rb
```diff
@@ -8,4 +8,8 @@ def self.passed?
   def self.iconize
     {class: :success, type: 'check-circle'}
   end
+
+  def self.exp_given
```
flbulgarelli, line 12: Why are we using `exp` instead of `experience`?

felipecalvo: `exp` and `xp` are common shorthands for `experience points` - you are not awarding experience, but experience points.

See https://en.wikipedia.org/wiki/Experience_point

I could change it to `experience_points_given` if needed...

flbulgarelli: ok!

felipecalvo:
This comment is messy so the merged code isn't.

- `UserStats` instead of `ExperiencePoints`: https://github.com/mumuki/mumuki-domain/pull/99/commits/74dc56c2dc533252a6202c4d955920feb2e60b51 overrides https://github.com/mumuki/mumuki-domain/pull/99/commits/a5fb75063ebf8c08ccb030e26df55793f6f526b5
- `Gamified` instead of `WithGamification`: https://github.com/mumuki/mumuki-domain/pull/99/commits/b3c6cf9d3dffd09d06dd1a91436d970e5489b3fc overrides https://github.com/mumuki/mumuki-domain/pull/99/commits/dedb765e8a2f0e4a46e9903e9737ec08a081613a
- No more comparing statuses with `<=>`: deleted https://github.com/mumuki/mumuki-domain/pull/99/commits/c0bf6b4aa78da515381580e3ff6ae75c502d377d
- `improved_by?` instead of `lesser_than?`: https://github.com/mumuki/mumuki-domain/pull/99/commits/f1734fdbf9cce2dc68c0d13c6fbfb8f22b881190 overrides https://github.com/mumuki/mumuki-domain/pull/99/commits/5d36a898a4a565bd6d7637a120676a29de9fb979. (I preferred `improved_by?`over `improves?`).
- Define `add_exp!`: https://github.com/mumuki/mumuki-domain/pull/99/commits/84c41ee6b6ad1a4c9b2548584dc30e177487cba6 overrides https://github.com/mumuki/mumuki-domain/pull/99/commits/1e46335ecadb5cd2b90e96084c3f1a186af4d1cd. (didn't go with `increment!` after flbulgarelli's warning)
- Move gamification stuff to `before_save` in Assignment: see https://github.com/mumuki/mumuki-domain/pull/99/commits/092ca5e0c2753d3c90a636470466a89d48f4bdea
- Make better gamification tests. I said I would merge them all in one file. I didn't. But now it more closely resembles the normal flow instead of using a magic helper. See https://github.com/mumuki/mumuki-domain/pull/99/commits/bfc35b88b76999b78c3cf3bb8d375487f1e7c4ad

PR #100, opened 2020-06-29
Use classroom as gem

luchotc:


luchotc:
Ready to review!

app/models/application_record.rb
```diff
@@ -95,7 +94,7 @@ def self.update_or_create!(attributes)
     obj
   end
 
-  def self.whitelist_attributes(a_hash, options={})
+  def self.whitelist_attributes(a_hash, options = {})
```
flbulgarelli, line 97: Side comment: maybe in the future we should consider this well-knoen discussion https://github.com/rails/rails/issues/33677

app/models/application_record.rb
```diff
@@ -60,9 +60,8 @@ def save_and_notify!
   end
 
   def update_and_notify!(data)
-    update! data
-    notify!
-    self
+    assign_attributes data
+    save_and_notify!
```
flbulgarelli, line 64: Does `save_and_notify` still return `self`, right?

spec/models/exam_spec.rb
```diff
@@ -7,18 +7,16 @@
 
   describe '#upsert' do
     let(:guide) { create(:guide) }
-    let(:exam_json) { {eid: '1', slug: guide.slug, start_time: 5.minutes.ago, end_time: 10.minutes.since, duration: 150, language: 'haskell', name: 'foo', uids: [], organization: 'test', course: course.slug, passing_criterion: {type: 'none'}} }
+    let(:exam_json) { {eid: '1', slug: guide.slug, start_time: 5.minutes.ago, end_time: 10.minutes.since, duration: 150, language: 'haskell', name: 'foo', uids: [], organization: 'test', course: course.slug, passing_criterion_type: 'none'} }
```
flbulgarelli, line 10: :open_mouth:  It is a subtle change the introduction of `passing_criterion_type`. May this have further consequences? May passing criterion also have payload data like the minimal grade?

PR #101, opened 2020-06-30
Adding gamification_enabled? flag

luchotc:
# :dart: Goal 

This PR adds a `gamification_enabled?` flag for organizations.

# :memo: Details

This flag is meant to define whether an organization is using gamification or not. 
This implies that this feature is opt in.


luchotc:
Resolves https://github.com/mumuki/mumuki-domain/issues/98

PR #102, opened 2020-07-06
Feature forum enhancements

luchotc:
This one:

- Fixes the discussion count that is fixed to the results by page currently
- Add a filter to allow ordering by useful_responses.
- It implicitly defines that a `useful` response is one which is given by a moderator or by a user but it's `verified` by a moderator.
- It de-normalizes some counts 
- It changes some filter orders by relevance 

To be done:
- Change `approved` semantic for something more meaningful like `verified` or `validated`

### DIFF
--- a/app/models/concerns/with_scoped_queries.rb
+++ b/app/models/concerns/with_scoped_queries.rb
@@ -16,8 +16,8 @@ module WithScopedQueries
   end
 
   class_methods do
-    def query_methods
-      queriable_attributes.keys
+    def query_methods(excluded_methods)
+      queriable_attributes.keys - excluded_methods.to_a
     end
 
     def scoped_query_module(method)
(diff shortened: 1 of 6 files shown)


app/models/concerns/with_scoped_queries.rb
```diff
@@ -28,19 +28,19 @@ def permitted_query_params
       queriable_attributes.values.flatten
     end
 
-    def actual_params(params, excluded_param)
-      params.reject { |it| it == excluded_param.to_s }
+    def actual_params(params, excluded_params)
+      params.except(*excluded_params)
     end
 
-    def scoped_query_by(params, excluded_param=nil)
```
flbulgarelli, line 35: Nice! Just be aware of compatibility issues

app/models/discussion.rb
```diff
@@ -131,6 +132,16 @@ def extra_preview_html
...
 
+  def update_counters_and_timestamps!
+    messages_query = messages.reorder(updated_at: :desc)
+    useful_messages = messages_query.select &:useful?
+    self.messages_count = messages_query.count
+    self.useful_messages_count = useful_messages.count
+    self.last_initiator_message_at = messages_query.find(&:from_initiator?)&.updated_at
+    self.last_moderator_message_at = useful_messages.first&.updated_at
+    save!
+  end
```
flbulgarelli, line 143: Why don't we use `update` here? It is cleaner and will just save those changes even when there are other dirty fields

luchotc:
Ready IMO! @fedescarpa

PR #103, opened 2020-07-06
Feature kindergarten

flbulgarelli:
# :dart: Goal

This PR introduces the `input_kindergarten` layout concept

# :memo: Details

I am converting the original `input_kids` into an "abstract" layout, and replacing it with `input_primary` and `input_kindergarten` while keeping the original `input_kids?` query method, and making the `layout=` setter to convert `input_kids` into `input_primary` for maximum compatibly.

# :beetle:  Bug fixes

Also, a minor bug - that prevented layout from been initialized in the `Problem` constructor -  has been fixed

# :back: Backward compatible

Aside of the previous considerations, this PR is **not** backward compatible, since it `layout` method will not answer `input_kids` anymore. 

# :eyes: Relations

See https://github.com/mumuki/mumuki-laboratory/pull/1420

### DIFF
--- a/app/models/exercise/reading.rb
+++ b/app/models/exercise/reading.rb
@@ -11,6 +11,10 @@ class Reading < Exercise
     false
   end
 
+  def layout=(layout)
+    raise 'can not set a layout different to input_bottom on readings' unless layout.like? :input_bottom
+  end
+
   def queriable?
     false
   end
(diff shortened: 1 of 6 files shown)


app/models/exercise/reading.rb
```diff
@@ -11,6 +11,10 @@ def input_kids?
     false
   end
 
+  def layout=(layout)
```
luchotc, line 14: Nice! We should make a list of layouts that are coupled to exercise types (if there are more) and generalize this one.

flbulgarelli: Yup. I want to remove this restriction instead of just documenting, actually

PR #106, opened 2020-07-17
Defining with_target_visual_identity mixin

### LINKED ISSUE #92: Segment avatars into age targets
An avatar should only be selectable on an organization which shares its intended age target.

For example, a kids organization would have kids avatars, adults would have their own set of avatars, etc.

That doesn't mean an user's selected avatar is organization-dependant - just the ones they are able to choose from.

luchotc:
# :dart: Goal

This PR introduces the `with_target_visual_identity` mixin
Resolves #92 

# :memo: Details

This one intends to be one of the two _target-related_ flags.
This one defines which will be the target visual identity of the organization and, in this case, the avatars.
Currently we have two very different aesthetics, the one for kids and the one for grown-ups, and this flag aims to allow to render certain resources based on the current one (which is determined by the current Organization).

The other flag, which isn't added here, is proposed to be `target_audience` which will define whether that organization is intended for `:kindergaten, :primary and (:secondary_and_above || :secondary, :adults)` -> I prefer the later.

This means that an organization may have a kids identity _visuallly_ but it is meant to be used by adults. This would be the case of an organization which uses kids exercises to teach adults programming basics.

  
# :back: Backward compatible

This PR is backward compatible, as it only adds new behavior. 

# :eyes: Reviewer questions

Would like to hear opinions about the two proposed flags.
In particular I'm not sure whether `target_visual_identity` is the best name and if we should use another word instead of `grown_ups`. I think that there isn't any doubt about `kids` as it matches the semantics introduced by @flbulgarelli.

# :heavy_plus_sign: Extra

Please see [my comment](https://github.com/mumuki/mumuki-domain/pull/106#pullrequestreview-450923955) bellow for extra information about these changes

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -31,12 +31,12 @@ class User < ApplicationRecord
   has_many :exams, through: :exam_authorizations
 
   enum gender: %i(female male other unspecified)
-
   belongs_to :avatar, optional: true
 
   before_validation :set_uid!
   validates :uid, presence: true
 
+  after_initialize :init
   PLACEHOLDER_IMAGE_URL = 'user_shape.png'.freeze
 
   resource_fields :uid, :social_id, :email, :permissions, :verified_first_name, :verified_last_name, *profile_fields
@@ -145,8 +145,12 @@ class User < ApplicationRecord
     exams.any? { |e| e.in_progress_for? self }
   end
 
+  def custom_profile_picture
+    avatar&.image_url || image_url
+  end
+
   def profile_picture
-    avatar&.image_url || image_url || placeholder_image_url
+    custom_profile_picture || placeholder_image_url
   end
 
   def bury!
@@ -180,7 +184,7 @@ class User < ApplicationRecord
   end
 
   def init
-    self.avatar ||= Avatar.sample_for(self)
+    self.avatar = Avatar.sample_for(self) unless custom_profile_picture.present?
   end
 
   def self.sync_key_id_field
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -173,8 +173,12 @@ class User < ApplicationRecord
     sequence[0..count + lookahead - 1]
   end
 
-  def current_visual_identity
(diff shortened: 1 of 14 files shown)


felipecalvo:
Va directamente en español: me quiero matar porque es exactamente lo que
estuve haciendo esta mañana jajaja.

De esto se aprende. Ahora lo reviso.

felipecalvo:
Ok, it's not so bad. I chose to have only a `target` flag: `kids` and `adults`, and though `grown-ups` sounded better at first, it seems they're synonyms. So we could go with whatever we like the most.

We can pick the last two commits at https://github.com/mumuki/mumuki-laboratory/commits/feature-show-target-avatars and add them to this PR.

By the way, I'm not sure why we'd (at least currently) want a `target_audience` enum. Do we have plans to care at all? Are there any non-visual aspects that would change for the different audiences?

app/models/user.rb
```diff
@@ -201,4 +209,16 @@ def self.configure_buried_profile!(profile)
   def self.buried_profile
     (@buried_profile || {}).slice(:first_name, :last_name, :email)
   end
+
+  def placeholder_image_url
+    PLACEHOLDER_IMAGE_URL
+  end
+
+  def current_organic_context
```
luchotc, line 217: I want to hear thoughts about this one too.
I think this can be generalized, specially with the introduction of `current_workspace`.
Although the `course` is not involved here, it might be.

luchotc:
> Ok, we worked on the same thing at the same time. (It's not so bad). I chose to have only a `target` flag: `kids` and `adults`, and though `grown-ups` sounded better at first, it seems they're synonyms. So we could go with whatever we like the most.
> 
> By the way, I'm not sure why we'd (at least currently) want a `target_audience` enum. Do we have plans to care at all? Are there any non-visual aspects that would change for the different audiences?

We could introduce a bit of custom logic based on that, as maybe it doesn't make sense to ask for login or track progress in certain cases, but it isn't defined yet.

PR #108, opened 2020-07-24
Feature serverless evaluation

flbulgarelli:
# :dart: Goal 

This PR allows clients to send an additional, non-persistable `client_result` param to `try_submit!`, which is the cornerstone of the serverless evaluation mode, that will allow browsers to propose evaluation results, which may be faster o better than server-side evaluation

# :memo: Notes

This PR just sends the new attributes to the bridge, which does not have any impact by itself. Runners may ignore the `client_result` completely. 

This PR is 100% backwards compatible

# :soon: Future work

In order to improve performance, further iterations of this evaluation strategy may even avoid hitting the runner. However, care should be taken that results are not tampered by malicious clients. , Trusting the `client_result` is currently a responsibility of the runner, but in the future, it may be **also** a responsability of the exercise - an exercise may have a `serverless_evaluation_enabled`, the organization - or workspace - or even a domain logic like `exam? => serverless evaluation disabled`.

### DIFF
--- a/app/models/concerns/submittable/solvable.rb
+++ b/app/models/concerns/submittable/solvable.rb
@@ -1,12 +1,16 @@
 module Solvable
-  def submit_solution!(user, attributes={})
-    assignment, _ = find_assignment_and_submit! user, attributes[:content].to_mumuki_solution(language)
+  def submit_solution!(user, submission_attributes={})
+    assignment, _ = find_assignment_and_submit! user, solution_for(submission_attributes)
     try_solve_discussions(user) if assignment.solved?
     assignment
   end
 
   def run_tests!(params)
     language.run_tests!(params.merge(locale: locale, expectations: expectations, custom_expectations: custom_expectations))
+  end
+
+  def solution_for(submission_attributes)
+    submission_attributes[:content].to_mumuki_solution(language)
   end
 end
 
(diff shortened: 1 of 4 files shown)


app/models/concerns/submittable/solvable.rb
```diff
@@ -1,13 +1,19 @@
...
   end
 
   def run_tests!(params)
     language.run_tests!(params.merge(locale: locale, expectations: expectations, custom_expectations: custom_expectations))
   end
+
+  def solution_for(submission_attributes)
+    submission_attributes[:content]
+      .to_mumuki_solution(language)
+      .with_client_result(submission_attributes[:client_result])
```
luchotc, line 15: Why don't we just add it to `to_mumuki_solution` method?

lib/mumuki/domain/submission/base.rb
```diff
@@ -9,6 +9,7 @@ class Mumuki::Domain::Submission::Base
                 :submission_id, :queries, :query_results, :manual_evaluation_comment]
 
   attr_accessor *ATTRIBUTES
+  attr_accessor :client_result
```
luchotc, line 12: Are we planning on using it in a submission type other than a `solution`?

flbulgarelli: I am not planning that, but that would be definitively a good idea to implement in the future. Playground and interactive exercises may benefit of this feature as well.

app/models/concerns/submittable/solvable.rb
flbulgarelli: That is because `to_mumuki_solution` is a polymorphic method, and doing so would repeat some code.

PR #109, opened 2020-07-30
Feature discussion creation flow

luchotc:


app/models/discussion.rb
```diff
@@ -67,7 +63,7 @@ def last_message_date
   end
 
   def friendly
-    title
+    initiator.name
```
luchotc, line 66: This is not ideal but I'm not sure if there's a better option now :confused:

spec/models/message_spec.rb
```diff
@@ -60,6 +60,7 @@
                            sender: 'teacher@mumuki.org',
                            read: false,
                            exercise: {bibliotheca_id: problem.bibliotheca_id},
+                           not_actually_a_question: false,
```
luchotc, line 63: We should except this one in `import_from_resource_h!` instead of adding it here

fedescarpa:
LGTM

PR #110, opened 2020-07-30
Email verification

julian-berbel:
This is what I've come up with so far :thinking: 

I'm thinking about the following behavior:
- If user's email is verified, then usual behavior
- If user's email is not verified and:
  - organization does not require verification, then usual behavior
  - organization demands immediate verification, user could be redirected somewhere asking to verify their email (1)
  - organization gives a grace period and:
    - user has not been notified yet, then redirect or show a modal asking to verify, but allow to continue (and start counting for grace period)
    - user has been notified and grace period is not over, then usual behavior
    - user has been notified and grace period is over, then same as (1)

(not sure the term `grace period` is correctly applied here, it's usually related to economic transactions)

policies' namespace seems obnoxiously long :thinking: 

anyways, wanted to check this all sounds about right so far @luchotc


julian-berbel:
Closing.
Implementing email verification in app doesn't really make sense. We should verify it in login provider for improved security.

PR #111, opened 2020-07-30
Adding the ability to restrict forum to just trusted users

luchotc:
# :dart: Goal

This PR allows certain organizations to restrict the forum for `trusted_for_forum` users. In the current implementation it's just a boolean but with the intention to be expanded with age validation, for example.
 
# :arrow_backward: Retro-compatibility

This PR is fully backwards compatible and should not affect any existing organization as the default is for users not to be trusted but the forum will be available for trusted and untrusted users by default.

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -105,17 +105,8 @@ class Organization < ApplicationRecord
   # ask for help in this organization
   #
   # Warning: this method does not strictly check user's permission
-  def ask_for_help_enabled?(user = nil)
-    report_issue_enabled? || community_link.present? || can_create_discussions?(user)
-  end
-
-  # Tells if the given user can
-  # create discussion in this organization
-  #
-  # This is true only when this organization has a forum and the user
-  # has the discusser pseudo-permission
-  def can_create_discussions?(user = nil)
-    forum_enabled? && (!user || user.discusser_of?(self))
+  def ask_for_help_enabled?(user)
+    report_issue_enabled? || community_link.present? || user.can_discuss_in?(self)
   end
 
   def import_from_resource_h!(resource_h)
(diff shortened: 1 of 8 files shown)


app/models/organization.rb
```diff
@@ -105,17 +105,8 @@ def site_name
   # ask for help in this organization
   #
   # Warning: this method does not strictly check user's permission
-  def ask_for_help_enabled?(user = nil)
```
luchotc, line 108: I'm removing this default as it comes from `assignment.submitter` and it can't be nil.

app/models/organization.rb
```diff
@@ -105,17 +105,8 @@ def site_name
...
-    report_issue_enabled? || community_link.present? || can_create_discussions?(user)
-  end
-
-  # Tells if the given user can
-  # create discussion in this organization
-  #
-  # This is true only when this organization has a forum and the user
-  # has the discusser pseudo-permission
-  def can_create_discussions?(user = nil)
-    forum_enabled? && (!user || user.discusser_of?(self))
```
luchotc, line 118: This line was odd. It made nil users able to create_discussions :eye:

flbulgarelli: Weird. It looks like it was on purpose, but I agree that this has little sense

db/migrate/20200730221001_add_trusted_for_forum_to_user.rb
```diff
@@ -0,0 +1,5 @@
+class AddTrustedForForumToUser < ActiveRecord::Migration[5.1]
+  def change
+    add_column :users, :trusted_for_forum, :boolean, default: false
```
flbulgarelli, line 3: Be aware this migration will update all the users in the table, which can take a lot of time. Maybe it will be a good idea to not add a default.

flbulgarelli:
PS: I like the fact that you are delegating more behaviour to user. This is a crucial need for the userless mode (PR pending)

db/migrate/20200730221001_add_trusted_for_forum_to_user.rb
luchotc: Nice catch :smile:

PR #112, opened 2020-08-04
Feature incognito mode

flbulgarelli:
# :dart: Goal 

Implement a polymorphic `Incognito` user - in a null-object style - and the foundations of an incognito mode. This concept is required for allowing a userless mode. 

# :memo: Details

* Added a `Mumuki::Domain::Incognito` singleton 
* Added an organization flag for enabling this mode - which is actually a laboratory task
* Added new delegations using a new naming schema when ambiguity may arise -  :busts_in_silhouette:`.*_at(`:books:`)` methods for user messages that take a content - as opposed to  :books:`.*_for(`:busts_in_silhouette:`)` methods that are content  messages that take an user as argument 

This PR should be 100% backwards compatible, since it does not remove any method nor it alters semantics

# :eyes: See also

https://github.com/mumuki/mumuki-laboratory/pull/1450

### DIFF
--- a/app/models/guide.rb
+++ b/app/models/guide.rb
@@ -52,8 +52,7 @@ class Guide < Content
                 on assignments.exercise_id = exercises.id
                 and assignments.submitter_id = #{user.id}
                 and assignments.submission_status = #{Mumuki::Domain::Status::Submission::Passed.to_i}").
-        where('assignments.id is null').
-        order('public.exercises.number asc')
+        where('assignments.id is null')
   end
 
   def first_exercise
--- a/app/models/guide.rb
+++ b/app/models/guide.rb
@@ -42,7 +42,7 @@ class Guide < Content
   end
 
   def next_exercise(user)
-    pending_exercises(user).order('public.exercises.number asc').first
+    user.next_exercise_at(self)
   end
 
   # TODO: Make use of pending_siblings logic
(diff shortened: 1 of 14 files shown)


app/models/guide.rb
```diff
@@ -52,7 +52,8 @@ def pending_exercises(user)
                 on assignments.exercise_id = exercises.id
                 and assignments.submitter_id = #{user.id}
                 and assignments.submission_status = #{Mumuki::Domain::Status::Submission::Passed.to_i}").
-        where('assignments.id is null')
+        where('assignments.id is null').
+        order('public.exercises.number asc')
```
luchotc, line 56: This ordering logic it's related to `next_exercise`, not to `pending_exercises`.
It's an important distinction because we'll use indicators to easily get pending_items but ordering logic is not included there.

flbulgarelli: Great, I had that doubt :smile:

flbulgarelli: Please review it again!

spec/models/incognito_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+
+  describe 'guide fooling' do
+    let(:guide) { create(:guide, exercises: [ create(:exercise), create(:exercise) ]) }
+
+    describe 'completion_percentage_for', organization_workspace: :test do
+      it { expect(guide.completion_percentage_for(user)).to eq 0 }
+    end
+
+    describe 'next_exercise', organization_workspace: :test do
+      # TODO this looks weird. Why do we need a non-polymorphic
```
luchotc, line 41: Currently we aren't using indicators for this one as progress is shared between organizations.
This change was made and had to be rolled back because of that.

PR #114, opened 2020-08-06
Chore forum general improvements

luchotc:
This PR together with https://github.com/mumuki/mumuki-laboratory/pull/1451 has one main goal: to improve performance on book discussions page. :top: 
It also addresses some minor issues :bug: :

- useful button wasn't being shown as `solved?` was delegated to `submission_status` instead of `discussion`. :+1: 
- a discussion with no moderator response but with another students answers was being marked as `requires_moderator_response: false`. :sos: 
- the discussions weren't being solved because of `reachable_statuses_for_initiator` method deletion. I'm adding it once again and hiding them visually. :rescue_worker_helmet: 
- two locales were mixed up between pt and en translations :england: :brazil: 
- approved is now displayed for non moderator users  :heavy_check_mark: 

Tests are added to ensure none of this cases happen again

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -20,6 +20,8 @@ class Assignment < Progress
   delegate :language, :name, :navigable_parent, :settings,
            :limited?, :input_kids?, :choice?, :results_hidden?, to: :exercise
 
+  delegate :completed?, :solved?, to: :submission_status
+
   alias_attribute :status, :submission_status
   alias_attribute :attempts_count, :attemps_count
 
--- a/app/models/book.rb
+++ b/app/models/book.rb
@@ -18,7 +18,7 @@ class Book < Content
   end
 
   def discussions_in_organization(organization = Organization.current)
-    Discussion.all.where(organization: organization).includes(exercise: [:language, :guide])
+    Discussion.where(organization: organization).includes(exercise: [:language, :guide])
   end
 
   def first_chapter
--- a/app/models/book.rb
+++ b/app/models/book.rb
@@ -8,8 +8,6 @@ class Book < Content
   has_many :complements, dependent: :destroy
 
   has_many :exercises, through: :chapters
-  has_many :discussions, through: :exercises
-  organic_on :discussions
 
   delegate :first_lesson, to: :first_chapter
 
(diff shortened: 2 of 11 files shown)


app/models/book.rb
```diff
@@ -19,6 +17,10 @@ def to_s
     slug
   end
 
+  def discussions_in_organization(organization = Organization.current)
```
luchotc, line 20: This one is odd. It will be addressed in another PR.
The controller should be organization_discussions instead of book_discussions, as it doesn't make sense now.
And this line could be removed.

app/models/book.rb
```diff
@@ -19,6 +17,10 @@ def to_s
...
+    Discussion.all.where(organization: organization).includes(exercise: [:language, :guide])
```
luchotc, line 21: Getting all the discussions for a book this way instead of getting all the exercises and getting all the discussions through there it's a **huge** performance improvement. Also, pre-fetching language and guide improves things too.

app/models/assignment.rb
```diff
@@ -20,6 +20,8 @@ class Assignment < Progress
   delegate :language, :name, :navigable_parent, :settings,
            :limited?, :input_kids?, :choice?, :results_hidden?, to: :exercise
 
+  delegate :completed?, :solved?, to: :submission_status
```
luchotc, line 23: This is behavior related to assignment, not to contextualization.

app/models/discussion.rb
```diff
@@ -105,10 +105,6 @@ def reachable_status_for?(user, status)
     reachable_statuses_for(user).include? status
   end
 
-  def allowed_statuses_for(user)
```
luchotc, line 108: Unused method!

app/models/book.rb
flbulgarelli: `Discussion.all.where` == `Discussion.where`?

app/models/concerns/submittable/solvable.rb
```diff
@@ -1,7 +1,7 @@
 module Solvable
   def submit_solution!(user, submission_attributes={})
     assignment, _ = find_assignment_and_submit! user, solution_for(submission_attributes)
-    try_solve_discussions(user) if assignment.solved?
+    try_solve_discussions!(user) if assignment.solved?
```
flbulgarelli, line 4: Great

spec/spec_helper.rb
```diff
@@ -51,7 +51,16 @@ def reindex_current_organization!
   reindex_organization! Organization.current
 end
 
+# ===============
+# Runner stubbing
+# ===============
+
+def stub_runner!(stubbed_response)
```
flbulgarelli, line 58: Nice!

PR #120, opened 2020-08-19
Chore user age

luchotc:


lib/mumuki/domain/extensions/time.rb
```diff
@@ -0,0 +1,5 @@
+class Time
+  def round_years_since(another_time)
+    (self.to_s(:number).to_i - another_time.to_s(:number).to_i) / 10e9.to_i
```
flbulgarelli, line 3: Why not `another_time.year - self.year`?

spec/models/user_spec.rb
```diff
@@ -325,4 +325,49 @@
...
+    end
+
+    context 'with birthdate' do
+      context 'in a leap year and the same month and day' do
+        before { user.birthdate = '08/12/2000' }
+        it { expect(user.age).to eq 20 }
+      end
+
+      context 'in a leap year one day before birthday' do
+        before { user.birthdate = '07/12/2000' }
```
flbulgarelli, line 348: We should always use ISO dates

lib/mumuki/domain/extensions/time.rb
luchotc: I'm using this one to get user age. 
We need to handle month and day as well.

spec/models/user_spec.rb
luchotc: True! I used this one as if it was a setting instead a field. My bad

lib/mumuki/domain/extensions/time.rb
luchotc: Further thoughts? Perhaps the name is not ideal, but I couldn't come up with a better one and I decided to move this logic here instead of having it in the user class

luchotc: @flbulgarelli I'm missing this one to integrate gamification into master :smile:

flbulgarelli: Ok. Go ahead.

However I think it can be more clear  and avoid that dangling `to_s` by using something like this, which is more intention revealing: 

```
(self.to_date -  another.to_date / 365).to_i
```

flbulgarelli: In the end we decided to leave the code as-is, since  - although it is hacky - it allows to properly deal with leap years.

Issue #118, opened 2020-08-10
Move indicator dirtiness up one level


luchotc:
Guides shouldn't be dirty. Making that calculation is not expensive at all and making this change will divide indicators recalculation impact between submission deliveries and content visualizations.

This will allow be useful for giving guide medals easily: https://github.com/mumuki/mumuki-domain/issues/117

PR #128, opened 2020-09-07
Feature medals

felipecalvo:
## :dart: Goal 

Add the concept of medal, which any content (book/guide/topic) can award upon completion.

## :soon: Note to reviewers

I went the way we talked with @luchotc and used indicators. However, it needs some more work, as any kind of dirtiness (either if content changed or if a previously passed solution is now failed) currently results in the medal being revoked.

I thought of adding an `achievements` relationship, or whatever we call it, to persist medals once achieved for an user. It would still use indicators, though, to see if any new medals were earned.

Another option would be adding a `once_completed?` boolean of sorts on Indicator, which is simpler and I like more, and would work similarly to `top_submission_status` on Assignment. It would _still_ revoke medals if the whole content was deleted, though, but that's a pretty unusual scenario. Lastly, it seems like it could be useful for other cases, unlike an `achievements` table.

Something I also have to add is support for knowing if a medal is earned right after a solution is sent, though this is not trivial to do without entirely ruining the indicators' performance (basically recalculating completion on every single submission), so I'm being cautious. We would have to enable it exclusively for gamified organizations or think of something more efficient, though it would also depend on how/when we'll be showing this information on laboratory, which we're not sure of yet.

## :soon: To be done shortlist

* ~Awarded medals shouldn't be revoked~ (see 6b064e3)
* ~Know if a medal was awarded on real-time~ this is a task for Laboratory, though _maybe_ something should be changed on indicators for gamified organizations
* ~More comprehensive tests, spanning multiple guides and topics~ (see 6351f71ed53c7a9472606eca9d6f85cb6082fb95)
* ~Add awarded medals to selectable avatars~ (see https://github.com/mumuki/mumuki-laboratory/pull/1469 and https://github.com/mumuki/mumuki-domain/pull/128/commits/bcaadcc7cd30d0ac246b28239bffb4ca19c13936)


luchotc:
I think both the `achievements` table and the `once_completed?` flag are valid options.
As you said, we should design taking into consideration which screens are going to need this information and how we should calculate it.
 
We have three main uses here:
1. After a submission, to know if that triggered a content completion.
2. In some screens which display content info, to communicate that certain medal can be achieved upon completion.
3. In the user profile, to display all medals (obtained and available).

1. I agree that we should be careful here not to destroy indicator's performance. [This one](https://github.com/mumuki/mumuki-domain/issues/118) is a key point in it. Also, I think we should consider not whether the organization is gamified or not, but if the indicator's related content has a medal. I'm not considering the first part to contemplate cases where the gamification is activated later on.
2. This comes for free due to first case. `user.medal_given_for? content` seems to do the trick, which delegates to whether the associated indicator is completed or not.
3. This one is a bit trickier and if we decide to show locked medals as well (which I think we should), `achievements` table won't be so effective, as it only covers won medals.
Here we could debate a bit more how to optimize the query. Thinking about it quickly maybe using: `Usage.where(organization: Organization.current).map(&:item).select(&:medal)` we could get all contents for a given organization which have an associated medal. Then we could iterate that list and ask if it was completed through the indicators association. Taking into account that this happens only in the user profile, I think it's a query we can live with.

What do you think about this? Am I missing some scenario? @felipecalvo

felipecalvo:
A few words on the status of this PR.

Two problems remain:

- We have no easy way of knowing the full list of unlockable medals for a certain organization. We thought of depending on indicators, but if an indicator doesn't exist (i.e. because no exercises have been solved for a certain guide) then there's no way to reach all content - except for `book + book.chapters + book.topics`, maybe, which I'd rather avoid.

For example, this test shows how there are only two unacquired medals - `another_guide_medal` is not taken into consideration because no exercises from that guide were solved.
https://github.com/mumuki/mumuki-domain/blob/bc132e1e91a96883d326c25089cbd63df44489bf/spec/models/awardee_spec.rb#L38-L45

- Some more changes to indicators on gamified organizations should be done so we can award medals on real time.

______________


Also, we spoke a long time about how indicators should have a `once_completed?` to avoid revoking medals, but `indicator.completed?` already works like that! See https://github.com/mumuki/mumuki-domain/blob/7082a83493eb64ba640c84542424dc27a8cb51b0/app/models/assignment.rb#L47 and https://github.com/mumuki/mumuki-domain/blob/7082a83493eb64ba640c84542424dc27a8cb51b0/app/models/assignment.rb#L59-L61

Of course, I only noticed this after I had implemented `once_completed?` myself :stuck_out_tongue:. Wrote some indicator tests to make sure it's the intended behaviour.

app/models/concerns/awardee.rb
```diff
@@ -0,0 +1,27 @@
...
+    current_indicators.partition(&:once_completed?)
+  end
+
+  def current_indicators
+    items_with_medals.map { |i| i.progress_for(self, Organization.current) }
+  end
+
+  def items_with_medals
+    Usage.where(organization: Organization.current).map(&:item).select(&:medal_id)
+  end
```
felipecalvo, line 26: The only (I believe) thing missing is how to improve this query, or if it at least we can avoid querying twice for acquired and unacquired medals. This might have to be solved through Laboratory - I'll add whatever I come up with on  https://github.com/mumuki/mumuki-laboratory/pull/1469

luchotc (approved):
Great @felipecalvo. Just minor details!
I think we can make *acquired_medals methods to only execute once on laboratory, but let's not worry about it just yet.

app/models/concerns/awardee.rb
```diff
@@ -0,0 +1,27 @@
...
+  def medals_for(indicators)
+    indicators.map { |i| i.content.medal }.compact
+  end
+
+  def all_medals
+    current_indicators.partition(&:once_completed?)
+  end
+
+  def current_indicators
+    items_with_medals.map { |i| i.progress_for(self, Organization.current) }
```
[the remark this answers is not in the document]
luchotc: Also let's please use: 
`@awardable_content_progress_here ||= ..` in order to avoid making the query twice to the database.

app/models/concerns/awardee.rb
```diff
@@ -0,0 +1,27 @@
...
+    medals_for all_medals.second
+  end
+
+  private
+
+  def medals_for(indicators)
+    indicators.map { |i| i.content.medal }.compact
+  end
+
+  def all_medals
```
luchotc, line 16: Maybe `achievable_medals_here`?

app/models/concerns/awardee.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  def unacquired_medals
+    @unacquired_medals ||= medals_for awardable_content_progress_here.reject(&:once_completed?)
+  end
+
+  private
+
+  def medals_for(progress)
+    Organization.current.gamification_enabled? ? progress.map { |i| i.content.medal } : []
```
flbulgarelli, line 13: Side note: I think it is always better to not hardcode `Organization.current` in code, but instead place an optional parameter organization with current as default code. That way we do not need to switch to an organization before using the method.

flbulgarelli: Long ago we actually write code like that and then we started using the described approache

felipecalvo: Sure :+1: I've done that myself, just didn't think of it this time

app/models/concerns/awardee.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  private
+
+  def medals_for(content)
+    content.map(&:medal)
+  end
+
+  def awardable_contents_in(organization)
+    @awardable_contents_in ||= organization.gamification_enabled? ? organization.all_contents.select(&:medal_id) : []
```
luchotc, line 17: We could move this to organization and make it just organization.awardable_contents_for(self). What do you think?

app/models/concerns/awardee.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  def completed_content_in(organization)
```
luchotc, line 20: `awarded_content_in` to reflect the semantics better?

app/models/organization.rb
```diff
@@ -134,6 +134,12 @@ def display_name
     name.gsub(/\W/, ' ').titleize
   end
 
+  def all_contents
+    Usage.where(organization: self)
+         .group_by(&:item_type)
+         .flat_map { |k, v| k.constantize.where(id: v.map(&:item_id)) }
```
luchotc, line 140: |item_type, items|* maybe?

felipecalvo:
All done. Waiting for @travisci's hamster wheels to spin again

PR #129, opened 2020-09-07
Feature progress transfer

julian-berbel:
I was trying to think of a way to extract the similar logic between the two operations which is pretty easy when thinking of an assignment (defining `copy_to!` as `dup.move_to!`) but it gets pretty intricate and has some undesired consequences when thinking of indicators and their children in particular.

I also realized when console testing this that it's pretty easy to end up with duplicate records, what should we do in those cases @luchotc? 

In terms of strategies, and in order of difficulty, I'm thinking:
~1- Not dealing with it and living with possible duplicates.~
2- Deleting progress for given content on destination and then copying.
3- Merging progress favoring source or destination.
4- All of the above, and letting transferer choose.

Needs tests.

Some thoughts that pop up while pondering about this
- parent of transferred item should be marked dirty in destination?
- actually, not dealing with duplicates [would probably cause problems](https://drive.google.com/file/d/1OpxjYR4eZfgFQPJ2GMBV8igSsTqmEBU_/view?usp=sharing) such as needing to check children are for different contents for an accurate `completion_percentage` and possibly assignments for the same guide not dirtying the same parent.
- how about transferring only orphaned assignments and letting indicator tree build itself?

### DIFF
--- a/app/models/progress.rb
+++ b/app/models/progress.rb
@@ -18,6 +18,7 @@ class Progress < ApplicationRecord
   end
 
   def transfer_to!(organization)
+    raise "Transferred progress' content must be available in destination!" unless content_available_in?(organization)
     relocate_on!(organization)
     save!
     delete_duplicates!
--- a/app/models/progress.rb
+++ b/app/models/progress.rb
@@ -17,16 +17,14 @@ class Progress < ApplicationRecord
     dup.transfer_to!(organization)
   end
 
-  def move_to!(organization)
-    transfer_to!(organization)
-  end
-
   def transfer_to!(organization)
     relocate_on!(organization)
     save!
     delete_duplicates!
     self
   end
+
+  alias_method :move_to!, :transfer_to!
 
   def relocate_on!(organization)
     assign_attributes organization: organization, parent: nil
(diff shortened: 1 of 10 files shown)


app/models/progress.rb
```diff
@@ -12,4 +12,42 @@ def parent
...
+    save!
+    self
+  end
+
+  alias_method :_move_to!, :transfer_to!
+
+  %i(copy_to! move_to!).each do |selector|
+    define_method(selector) do |organization|
+      raise "Transferred progress' content must be available in destination!" unless content_available_in?(organization)
+      dirty_parent_by_submission!
```
julian-berbel, line 32: this is actually only necessary when moving progress, not when copying

spec/models/progress_spec.rb
```diff
@@ -3,27 +3,61 @@
...
+
+  # Empty book to test content validation on destination
+  let(:empty_book) { create :book }
+
+  # Main content tree
+  #     B
+  #     |
+  #     T
+  #   /  \
+  #  G1   G2
```
flbulgarelli, line 18: :heart:

PR #130, opened 2020-09-08
Adding required methods for incognito

luchotc:


lib/mumuki/domain/incognito.rb
```diff
@@ -88,6 +92,10 @@ def _read_attribute(key)
       raise "unknown attribute #{key}"
     end
 
+    def new_record?
```
luchotc, line 95: receiving `persisted?` in after_init callback

PR #134, opened 2020-09-16
Chore top submission status as app default

luchotc:


felipecalvo (approved):
I wish we didn't have to do this :rofl: 

Just for the record, you found this seems to be solved on Postgre 11, as the initialization is done at query time instead of preemptively.

!RemindMe when we have Postgre 11

julian-berbel (approved):
makes sense!

PR #138, opened 2020-10-06
Fix dead indicator tree branches

julian-berbel:
These tests prove the behavior we were talking about the other day @luchotc.

As for a fix, I'm thinking we could just add a `before_destroy` to `Usage` and either delete, or unlink associated the corresponding indicators. Any preference regarding that? 

Deleting is the better option if the content is never readded to the organization, since we'd be keeping less useless records (although we don't delete assignments).
If we expect the content to be readded to the organization later then unlinking would be better as the relinking process would be less resource intensive (we'd only have to relink indicators instead of assignments, which are fewer in numbers).

### DIFF
--- a/app/models/indicator.rb
+++ b/app/models/indicator.rb
@@ -57,6 +57,10 @@ class Indicator < Progress
     children_passed_count == children_count
   end
 
+  def self.delete_all_for!(content, organization)
+    where(content: content, organization: organization).delete_all
+  end
+
   private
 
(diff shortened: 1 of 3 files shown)


luchotc:
Based on what has happened so far, let's delete it. I can't recall a single case in which we deleted a content and then re-added it.

luchotc (approved):
Awesome :heart:

PR #139, opened 2020-10-09
Feature add audit fields to discussion status changes

flbulgarelli:
# :dart: Goal 

In a lot of contexts @NadiaFinzi @Gustrucco and @rgonzalezt have pointed out that it would be nice to track not only when a moderator posts a message, but also when a moderator solves or closes a discussion. This PR adds audit fields to status changes,. 

# :beetle: Potential bugs found

I have noticed that there are already similar audit fields for last_access, but FK to user is stored as a string ( :exclamation: ) and without and index (:thinking:): 

```ruby
add_column :discussions, :last_moderator_access_by_id, :string
add_column :discussions, :last_moderator_access_at, :datetime
```
 
I am not sure if that was done as intended or not, but I think it is weird. I am not changing this as part of this PR, though. 

# :mag: Notes for reviewer 

I have little experience with the discussion model, so a highly critical review would be appreciated. 

This is new feature, but it is backwards compatible. Not changes in client code - aside of migrations - are required.

### DIFF
--- a/spec/models/discussion_spec.rb
+++ b/spec/models/discussion_spec.rb
@@ -89,6 +89,7 @@ describe Discussion, organization_workspace: :test do
         before { discussion.update_status!(:solved, moderator) }
 
         it { expect(discussion.status).to eq :solved }
+        it { expect(discussion.reachable_statuses_for initiator).to eq [] }
         it { expect(discussion.status_updated_by).to eq moderator }
         it { expect(discussion.status_updated_at).to be < Time.now }
         it { expect(discussion.reachable_statuses_for moderator).to eq [:opened, :closed] }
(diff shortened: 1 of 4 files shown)


spec/models/discussion_spec.rb
```diff
@@ -87,7 +89,8 @@
         before { discussion.update_status!(:solved, moderator) }
 
         it { expect(discussion.status).to eq :solved }
-        it { expect(discussion.reachable_statuses_for initiator).to eq [] }
```
flbulgarelli, line 90: I think this line was wrong, so I removed it.

luchotc (approved):
Niceee @flbulgarelli :tada: 
You are right. That it's in a fact a bug! Do you want to add it to this one or do you prefer to open a separate issue for that?

PR #140, opened 2020-10-13
Feature submit to main organization

flbulgarelli:
# :dart: Goal

To avoid the need of most resubmissions by sending to the main immersive organization if present

# :memo: Details

This PR introduces the concept of `immersive_organization_at` which is the place where user is _supposed_ to be in _most_ cases. This organization is determined using explicit permissions, content sharing, target audience, publicity and immersivity. 

This PR also extends the concept of `current_organic_context` to `current_immersive_context`, which supersedes it by  taking immersivity into account. 

Using those concepts, the following notification-specific actions are taken: 

1. Assignments that belong to more than a single context are notified to all of them
2. Such assignments are also marked as _missplaced_ reflecting that they are possible in the wrong context
 

# :soon: Future work

This PR assumes that in the future the UI may actually redirect user to such organic contexts. In that scenario, notifying to all of them will be useless since that scenario may not happen anymore. However the building blocks of `current_immersive_context_at` and `immersive_organization_at` will be still meaningful. 

# :thinking: Doubts

I am still considering that a user's current organization is immersed into another when she has explicit **student** permissions. This is an old design decision, but perhaps we could extend it to teachers, too.

I am not notifying the new `misplaced` flag, but perhaps it may be a good idea to do so. 

# :ballot_box_with_check: Pending work

I should test the misplaced flag, too.

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -91,7 +91,13 @@ class Assignment < Progress
   end
 
   def notify!
-    Mumukit::Nuntius.notify! 'submissions', to_resource_h unless Organization.silenced?
+    unless Organization.silenced?
+      notification_contexts.uniq.each { |it| Mumukit::Nuntius.notify! 'submissions', to_resource_h(context: it) }
+    end
+  end
+
+  def notification_contexts
+    [Organization.current, submitter.current_immersive_context_at(exercise)]
   end
 
   def notify_to_accessible_organizations!
@@ -167,7 +173,8 @@ class Assignment < Progress
     exercise.run_tests! params.merge(extra: extra, test: test, settings: settings)
   end
 
-  def to_resource_h
+  def to_resource_h(options = {})
+    context = options[:context] || Organization.current
     excluded_fields = %i(created_at exercise_id id organization_id parent_id solution submission_id
                          submission_status submitted_at submitter_id top_submission_status updated_at)
 
@@ -182,7 +189,7 @@ class Assignment < Progress
                 exercise: {only: [:name, :number]},
                 submitter: {only: [:email, :social_id, :uid], methods: [:name, :profile_picture]}}).
       deep_merge(
-        'organization' => submitter.current_immersive_context_at(exercise).name,
+        'organization' => context.name,
         'sid' => submission_id,
         'created_at' => submitted_at || updated_at,
         'content' => solution,
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -182,7 +182,7 @@ class Assignment < Progress
                 exercise: {only: [:name, :number]},
                 submitter: {only: [:email, :social_id, :uid], methods: [:name, :profile_picture]}}).
       deep_merge(
-        'organization' => submitter.recommended_organic_context_at(guide).name,
+        'organization' => submitter.recommended_organic_context_at(exercise).name,
         'sid' => submission_id,
         'created_at' => submitted_at || updated_at,
         'content' => solution,
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -94,12 +94,12 @@ class Assignment < Progress
     unless Organization.silenced?
       contexts = current_notification_contexts
       update_missplaced!(contexts.size > 1)
-      contexts.uniq.each { |it| Mumukit::Nuntius.notify! 'submissions', to_resource_h(context: it) }
+      contexts.each { |it| Mumukit::Nuntius.notify! 'submissions', to_resource_h(context: it) }
     end
   end
 
   def current_notification_contexts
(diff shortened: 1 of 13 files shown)


julian-berbel:
could a user possibly be a member of 2 immersive organizations?

flbulgarelli:
> could a user possibly be a member of 2 immersive organizations?

Yes, it could happen. Because of that I have modified this behavior to only occur when there is

app/models/assignment.rb
```diff
@@ -91,7 +91,15 @@ def extension
   end
 
   def notify!
-    Mumukit::Nuntius.notify! 'submissions', to_resource_h unless Organization.silenced?
+    unless Organization.silenced?
+      contexts = notification_contexts
+      update_missplaced!(notification_contexts.size > 1)
```
flbulgarelli, line 96: typo :facepalm:

julian-berbel:
> > could a user possibly be a member of 2 immersive organizations?
> 
> Yes, it could happen. Because of that I have modified this behavior to only occur when there is only a single result @julian-berbel :
> 
> ```ruby
>  def immersive_organization_at(path_item, current = Organization.current)
>     return nil unless current.immersible?
> 
>     usage_filter = path_item ? lambda { |it| path_item.used_in?(it) } : lambda { true }
>     student_granted_organizations
>       .select { |it| current.immersed_in?(it) }
>       .select(&usage_filter)
>       .single
>   end
> ```

Understandable, have a great day

app/models/assignment.rb
```diff
@@ -91,7 +91,15 @@ def extension
...
+      contexts.uniq.each { |it| Mumukit::Nuntius.notify! 'submissions', to_resource_h(context: it) }
+    end
+  end
+
+  def notification_contexts
```
flbulgarelli, line 101: I should rename this to `current_notification_contexts` just for naming consistency

app/models/assignment.rb
```diff
@@ -167,9 +175,10 @@ def run_tests!(params)
     exercise.run_tests! params.merge(extra: extra, test: test, settings: settings)
   end
 
-  def to_resource_h
+  def to_resource_h(options = {})
```
flbulgarelli, line 178: I should test this change, too

lib/mumuki/domain/helpers/user.rb
```diff
@@ -87,8 +87,21 @@ def main_organization
     student_granted_organizations.first || any_granted_organizations.first
   end
 
+  # Deprecated: use `immersive_organization_at` which
+  # properly looks for a single immersive organization taking
+  # current organization and path into account
   def has_immersive_main_organization?
-    !!main_organization.try(&:immersive?)
+    main_organization.try { |it| it if it.immersive? }.present?
```
luchotc, line 94: This could be `main_organization.try(&:immersive?).present?`, right?
`nil` and `false` both evaluates to `false` for `.present?`

spec/models/organization_spec.rb
```diff
@@ -262,4 +262,22 @@
       it { expect(organization.display_name).to be_nil }
     end
   end
+
+  describe '#immersible?' do
+    context 'current public' do
+      let(:one) { build(:organization, public: true) }
+
+      let(:other_immersive) { build(:organization, immersive: true) }
+      let(:other_non_immersive) { build(:organization, immersive: true) }
```
luchotc, line 271: I guess this one is meant to be false :thinking:, right? And the test below also has the wrong value

app/models/assignment.rb
luchotc: As we aren't using the options for anything else it would be cleaner to use a named param with Organization.current as default, right?

lib/mumuki/domain/helpers/user.rb
```diff
@@ -87,8 +87,21 @@ def main_organization
...
+  end
+
+  def immersive_organization_at(path_item, current = Organization.current)
```
luchotc, line 97: In this PR I don't see where path_item could be nil. What other cases are you considering?

lib/mumuki/domain/helpers/user.rb
```diff
@@ -87,8 +87,21 @@ def main_organization
...
+    return nil unless current.immersible?
+
+    usage_filter = path_item ? lambda { |it| path_item.used_in?(it) } : lambda { |_| true }
+    student_granted_organizations
+      .select { |it| current.immersed_in?(it) }
+      .select(&usage_filter)
+      .single
```
luchotc, line 104: This seems a good choice to avoid extreme duplication of data!

luchotc (approved):
This is great :heart:

flbulgarelli:
Closing. See #145

PR #141, opened 2020-10-16
Convert invalid lookahead values into nil

flbulgarelli:
# :dart: Goal

Avoid having invalid values in organization's `progressive_display_lookahead`

# :memo: Details

`progressive_display_lookahead` must be nil or positive, but not negative or zero. This PR just silently corrects invalid values. I could have just validated that condition and raise an error, but I think this behavior is more  API-friendly and still non-surprising.

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -135,7 +135,7 @@ class Organization < ApplicationRecord
   end
 
   def progressive_display_lookahead=(lookahead)
-    self[:progressive_display_lookahead] = lookahead && lookahead > 0 ? lookahead : nil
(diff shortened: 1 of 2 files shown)


app/models/organization.rb
```diff
@@ -134,6 +134,10 @@ def display_name
     name.try { |it| it.gsub(/\W/, ' ').titleize }
   end
 
+  def progressive_display_lookahead=(lookahead)
+    self[:progressive_display_lookahead] = lookahead && lookahead > 0 ? lookahead : nil
```
luchotc, line 138: You could use lookahead&.positive? :smile:

luchotc (approved):
Just a detail which you can ignore :smiley:

PR #146, opened 2020-10-22
Do not sanitize affable test results

julian-berbel:
Fixes some result display problems where something like:
```
Expected: <49> but got <50>
```

would show up as:
```
Expected: &lt;49&gt; but got &lt;50&gt;
```

The problem is not really in the sanitization itself, but on the results rendering flow:
1- We sanitize results, turning ` ```<49>``` ` into ` ```&lt;49&gt;``` ` (not as a result of sanitization but rather as a result of parsing as html fragment and converting back to string).
2- We then convert markdown to html which results in `<code>&lt;49&gt;</code>`.

The problem with that is that the content within the `code` tags is not displayed as html, so the `&lt;` and `&gt;` are not rendered as `<` and `>`.

Removing step 1 fixes the visual issue, and with the current it has does not introduce a security issue as we're then processing results with `test_result_html`, which sanitizes the resulting html.

We _should_ however be aware that this means affable_test_results can include malicious code which should be sanitized at the appropiate time.

### DIFF
--- a/app/models/concerns/contextualization.rb
+++ b/app/models/concerns/contextualization.rb
@@ -135,4 +135,8 @@ module Contextualization
           status: it[:status])
     end
   end
+
+  def sanitized_affable_test_results
+    affable_test_results.each { |it| it[:result] = it[:result]&.sanitized }
+  end
 end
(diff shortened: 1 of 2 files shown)


flbulgarelli:
Hi! In the end, will we go for this approach or will we try to fix at  sanitization/nokogiri level?

If we are going for the first option, then let's add a test and merge this one

luchotc:
See: https://github.com/mumuki/mumuki-laboratory/pull/1490

luchotc:
:+1: to @flbulgarelli comment

Issue #147, opened 2020-10-23
Decouple platform from central concept


flbulgarelli:
The concept of a `central` organization is long-lasting, and was tied to original `mumuki.io` service in 2016. 

However, both platform and service have evolved far from it, and the requierment of having a central organization does not exist anymore - now we have better concepts like public organizations and [the forthcoming `immersible` flag](https://github.com/mumuki/mumuki-domain/issues/144). 

Now it is time to seek for those usages and remove them.

PR #148, opened 2020-10-27
Add display and page organization fields

flbulgarelli:
# :dart: Goal

Introduce new `display` fields that are improved versions of `name` and `description` organization fields, and `page` accessors that act as a bridge between classic organizations that display book descriptive fields, and a new generation of organizations that can display their own data. 

# :memo: Details

* Introduced a `teaser_on` macro for generation markdown teasers
* Introduced `page` methods in organization
* Introduced `display` methods in organization, keeping compatibility with the preexisting `display_name`
* Some olds methods deprecated - see https://github.com/mumuki/mumuki-domain/commit/0f1357252effad4dfaeea4bf05284a3b4810bd5d

# :eyes: See also

* https://github.com/mumuki/mumuki-laboratory/pull/1496
* https://github.com/mumuki/mumuki-domain/issues/147

### DIFF
--- a/app/models/concerns/with_description.rb
+++ b/app/models/concerns/with_description.rb
@@ -2,11 +2,8 @@ module WithDescription
   extend ActiveSupport::Concern
 
   included do
-    markdown_on :description, :description_teaser, skip_sanitization: true
+    markdown_on :description, skip_sanitization: true
+    teaser_on :description
     validates_presence_of :description
-  end
-
-  def description_teaser
-    description.markdown_paragraphs.first
   end
 end
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -3,6 +3,16 @@ class ApplicationRecord < ActiveRecord::Base
 
   delegate :whitelist_attributes, to: :class
 
+  def self.teaser_on(*args)
+    args.each do |selector|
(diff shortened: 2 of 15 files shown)


app/models/concerns/with_description.rb
```diff
@@ -2,11 +2,8 @@ module WithDescription
   extend ActiveSupport::Concern
 
   included do
-    markdown_on :description, :description_teaser, skip_sanitization: true
+    markdown_on :description, skip_sanitization: true
```
luchotc, line 5: We could get rid of this one, right?

luchotc (approved):
image

app/models/concerns/with_description.rb
flbulgarelli: Do you mean, the  `description` or the `teaser`?

flbulgarelli: I did a quick search and it looks like they are being used in laboratory.. I didn't perform a deep search, though

luchotc: Oh my bad, I thought it said description_teaser which is being defined in teaser_on. Never mind me :sweat:

julian-berbel (approved):
Looks good!

PR #149, opened 2020-10-27
Chore: add immersible flag to organization

### LINKED ISSUE #144: Add immersible flag to organization
See https://github.com/mumuki/mumuki-domain/pull/140. We are using `immersible?` as an alias for `public?`. We need to add that one as a separate flag, as not all public organizations will be immersible. We should validate that an organization can't be immersible and immersive at the same time.

faloi:
Closes #144 

# :dart: Goal

Add immersible flag to organizations, replacing old `public?` alias.

# :memo: Details

* Flag is false by default.
* Test organization was made immersible, to preserve retrocompatibility.

# :warning: Considerations

After this is merged, all the public organizations will not be immersible anymore. Perhaps the code to make the selected ones immersible could be included in the migration.

### DIFF
--- a/db/migrate/20201027134205_add_immersible_to_organization.rb
+++ b/db/migrate/20201027134205_add_immersible_to_organization.rb
@@ -1,5 +1,5 @@
 class AddImmersibleToOrganization < ActiveRecord::Migration[5.1]
   def change
-    add_column :organizations, :immersible, :boolean, default: false
+    add_column :organizations, :immersible, :boolean
   end
 end
--- a/db/migrate/20201027134205_add_immersible_to_organization.rb
+++ b/db/migrate/20201027134205_add_immersible_to_organization.rb
@@ -1,0 +1,5 @@
+class AddImmersibleToOrganization < ActiveRecord::Migration[5.1]
+  def change
+    add_column :organizations, :immersible, :boolean, default: false
+  end
+end
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -147,7 +147,7 @@ describe User, organization_workspace: :test do
   end
 
 
-  describe 'immeresive behaviour' do
+  describe 'immersive behaviour' do
     let(:student) { create :user }
 
     context 'when no granted organizations' do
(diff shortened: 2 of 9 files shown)


db/migrate/20201027134205_add_immersible_to_organization.rb
```diff
@@ -0,0 +1,5 @@
+class AddImmersibleToOrganization < ActiveRecord::Migration[5.1]
+  def change
+    add_column :organizations, :immersible, :boolean, default: false
```
flbulgarelli, line 3: Because of migration performance reasons, we tend to avoid to add default boolean values, at least in big tables like users and assignments. But it is ok to add it for organizations, right @luchotc ?

flbulgarelli:
Looks good to me

spec/models/user_spec.rb
```diff
@@ -147,7 +147,7 @@
   end
 
 
-  describe 'immeresive behaviour' do
+  describe 'immersive behaviour' do
```
flbulgarelli, line 150: xD

db/migrate/20201027134205_add_immersible_to_organization.rb
flbulgarelli: I removed the default, just for consistency with the other fields

spec/models/user_spec.rb
faloi: Classic Franco.

luchotc (approved):
Niceee!
We aren't using `immersible?` for any core behavior so it's ok to deploy this one and then run the update afterwards.

A validation that checks that an organization can't be immersive and immersible at the same time will be nice to avoid undesired behavior.

flbulgarelli:
> A validation that checks that an organization can't be immersive and immersible at the same time will be nice to avoid undesired behavior.

@faloi could you take that one?

faloi:
> > A validation that checks that an organization can't be immersive and immersible at the same time will be nice to avoid undesired behavior.
> 
> @faloi could you take that one?

Sure. Any recommendation for the implementation? I guess a validation hook or something already present.

flbulgarelli:
> Sure. Any recommendation for the implementation? I guess a validation hook or something already present.

I'd use `validate` with a validation method: https://guides.rubyonrails.org/active_record_validations.html#custom-methods

flbulgarelli:
I migrated ci to travis-ci.com. It is working but it is not being reflected. I will investigate later. 

Also, for the future, please always remember to add pt and es-CL i18n.

PR #151, opened 2020-10-29
Fix allow strings on progressive display lookahead

julian-berbel:


luchotc (approved):
Ohhh nice, nil ends up being nil anyway. :100:

PR #152, opened 2020-10-30
Add some methods for immersive context

faloi:
Related to mumuki/mumuki-laboratory#1489

This adds some methods that will be used on laboratory.

### DIFF
--- a/lib/mumuki/domain/incognito.rb
+++ b/lib/mumuki/domain/incognito.rb
@@ -48,6 +48,14 @@ module Mumuki::Domain
       false
     end
 
+    def current_immersive_context_at(_)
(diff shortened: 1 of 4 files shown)


lib/mumuki/domain/incognito.rb
```diff
@@ -48,6 +48,14 @@ def can_access_teacher_info_in?(*)
       false
     end
 
+    def current_immersive_context_at(_)
```
flbulgarelli, line 51: :clap:

flbulgarelli: Kudos for considering the incognito mode

PR #155, opened 2020-10-31
Set parent_id as nil if assignment is recontextualized

luchotc:
This is probably not the cleanest solution but something like this is necessary because if we change assignments organization but we don't change its parent it remains attached to it and progress is never recalculated properly for that indicator

### DIFF
--- a/lib/mumuki/domain/submission/base.rb
+++ b/lib/mumuki/domain/submission/base.rb
@@ -54,6 +54,7 @@ class Mumuki::Domain::Submission::Base
   def save_submission!(assignment)
     assignment.content = content
     if assignment.organization != Organization.current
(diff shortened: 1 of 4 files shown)


luchotc:
@julian-berbel What do you think about this one?
Of course it needs a test proving that behavior, but I'm pretty sure that re-contextualization without emptying its parent leads to an inconsistency in indicators.

julian-berbel:
absolutely, that being said, for consistency's sake we should also dirty the indicator before moving the assignment

luchotc:
@julian-berbel True that! 
Could I delegate to you adding that line and a quick test that proves that behavior? :smile:

luchotc:
Great @julian-berbel !

PR #156, opened 2020-10-31
Use evaluation mode in solutions only

flbulgarelli:
# :dart: Goal

To make evaluation mode to be meaningful in solutions only. 

# :beetle: Issues solved

* Fixes https://github.com/mumuki/mumuki-laboratory/issues/1413


flbulgarelli:
PS: just for the record, there there are some captures

image

image

PR #158, opened 2020-11-04
Chore terms

luchotc:
# Goal :goal_net: 

To add localized, easily modifiable platform terms


lib/mumuki/domain/incognito.rb
```diff
@@ -56,6 +56,26 @@ def immersive_organizations_at(_)
...
+    def accepted_profile_terms?
+      true
+    end
+
+    def accepted_forum_terms?
+      true
+    end
+
+    def has_accepted?(term)
+      false
```
flbulgarelli, line 76: Shouldn't this be `true` just for sake of consistency?

_The following discussion is just for the future :clock1: _

I think for not it will be OK assuming an incognito doesn't need to accept anything - since this is an experimental feature -, but for the long term I'd prefer forcing incognitos to accept them too. 

For instance, accepting cookies policies is mandatories even in sites that don't require login.

flbulgarelli (approved):

lib/mumuki/domain/incognito.rb
luchotc: I'll change right ahead. The previous implementation was needs_acceptance?(term) so the refactor left that false forgotten :laughing:

luchotc: Nevermind my old comment, I did change it. This doesn't determine if it needs acceptance, it tell us if it user has_accepted that. So a false value will show all the terms open as if were never accepted in the first place.

The behavior works as you expected

PR #160, opened 2020-11-10
Allow saving empty solution to prevent inconsistencies

julian-berbel:


luchotc (approved):

app/models/assignment.rb
```diff
@@ -118,7 +118,7 @@ def self.evaluate_manually!(teacher_evaluation)
   end
 
   def content=(content)
-    if content.present?
+    unless exercise.reading?
       self.solution = exercise.single_choice? ? exercise.choice_index_for(content) : content
```
flbulgarelli, line 122: That is a huge change! Can't this have subtle unexpected effects? Is this code  fixing an issue?

julian-berbel: it's not really so much fixing an issue as making one more transparent, and also lessening its effects.
Right now, if a user has a passed assignment on an exercise and sends an empty solution, the tests are run and the results are saved, while the solution's content is not, leading to inconsistencies.

luchotc: We could make this PR more clear by adding a test prior to this change in which we show that behavior. What do you think @julian-berbel?

julian-berbel: sounds good!

luchotc:
For me it's ready but i want to hear @flbulgarelli further thoughts

luchotc:
@flbulgarelli has orally approved this one :smile: 
We can rebase it and merge it @julian-berbel

flbulgarelli:
> @flbulgarelli has orally approved this one smile

To sum up, the original code was so wrong that I did don't understand how I could think it was right :stuck_out_tongue:

PR #161, opened 2020-11-10
Adding role term acceptance methods

luchotc:
## Goal

Adding `has_role_terms_to_accept?` to check if the user needs to accept some role terms.
Also, it simplifies incognito terms related methods


flbulgarelli (commented):
Looks good, but perhaps some tests would be nice

PR #164, opened 2020-11-17
Chore recontextualize

luchotc:


julian-berbel (approved):
Looks great!

app/models/assignment.rb
```diff
@@ -52,6 +52,14 @@ def set_current_organization!
     self.organization = Organization.current
   end
 
+  def recontextualize!
```
flbulgarelli, line 55: I know this code predates this PR, but I'd suggest adding an organization param to make it independent of current organization

luchotc: This is a great suggestion! I've added it in https://github.com/mumuki/mumuki-domain/commit/163759388b3f070877ea4cbf956174f03545fa79
Hopefully this code will be gone shortly. This kind of automatic recontextualization is only necessary due to assignment sharing.

PR #170, opened 2020-12-11
Add mono_lesson? method to topic

fedescarpa:


app/models/topic.rb
```diff
@@ -35,6 +35,10 @@ def reindex_usages!
     Chapter.where(topic: self).map(&:book).each(&:reindex_usages!)
   end
 
+  def mono_lesson?
+    lessons.size == 1
```
luchotc, line 39: Prefer count as it's an AR relation. I'll change it :smile:

flbulgarelli:
Looks good!

My only comment is that I expected `monolesson` since its a single word, not two - `mono-` is just a prefix as in `monorail` or `monolithic`

luchotc:
@flbulgarelli I could use a re-review :smile:

app/models/chapter.rb
```diff
@@ -13,6 +13,9 @@ class Chapter < ApplicationRecord
 
   has_many :exercises, through: :topic
 
+  delegate :monolesson?, :monolesson, :first_lesson, to: :topic
+
+  delegate :next_exercise, :stats_for, to: :monolesson, allow_nil: true
```
flbulgarelli, line 18: Nice

flbulgarelli (approved):

PR #172, opened 2020-12-18
Feature randomize expectations

flbulgarelli:
# :dart: Goal

To allow `expectations` and `custom_expectations` to be randomized. 

# :memo: Details

This PR was tougher than I thought, since those fields also interact with the guide content fields inheritance. Thus, I implemented the following changes: 

1. Extended randomization to arrays and hashes in a polymorphic way. I also decided to treat symbols and nils the same way instead of current type-specific considerations
2. I had to make randomization compatible with fields inheritance by using `Problem`-specific aliasing instead of reading the raw fields. 
3. I extended and tested the `own_` fields in order to better reflect these semantic, and renamed the old `own_expectations` to `raw_expectations`.
4. I dropped original the `own_custom_expectations` polymorphic method from `Guide`, since it was never actually used. 
5. I dropped `guide_expectations` and `guide_custom_expectations` because of the same reason. 
5. I tidy up some extension methods

This change should be backwards compatible as long as my previously stated assumptions are true.


spec/models/guide_spec.rb
```diff
@@ -98,6 +98,30 @@
...
+    end
+
+    let(:exercise) do
+      build(:exercise,
+        extra: 'foo exercise extra',
+        expectations: [{binding: 'foo', inspection: 'Not:Uses:foo'}],
+        custom_expectations: 'expect: assigns foo;')
+    end
+
+    it { expect(guide.raw_expectations).to eq guide.raw_expectations }
```
felipecalvo, line 117: I think you meant this? Or am I missing something?

```suggestion
    it { expect(guide.raw_expectations).to eq [{binding: 'guide', inspection: 'Uses:expectations'}]}
```

app/models/exercise/problem.rb
```diff
@@ -21,20 +21,15 @@ def reset!
     self.expectations = []
   end
 
+  alias_method :own_expectations, :expectations
+  alias_method :own_custom_expectations, :custom_expectations
```
luchotc, line 25: I was about to suggest the use of `revamp_accessor` here but it's better that we keep the original behavior in `own_custom_expectations`

spec/models/guide_spec.rb
luchotc: :+1:

luchotc (approved):
Everything looks great! Let's wait for @julian-berbel feedback and merge it

app/models/exercise/challenge.rb
```diff
@@ -8,8 +8,10 @@ def reset!
     self.layout = self.class.default_layout
   end
 
+  alias_method :own_extra, :extra
+
   def extra(*)
-    [guide.extra, super]
+    [guide.extra, own_extra]
```
julian-berbel, line 14: I'm guessing that's the point, but this call changes as it's not passing arguments, is that alright?

luchotc: Actually this is a great point. I can't see any method call to extra which take a param. Is that splat needed altogether?

app/models/exercise/challenge.rb
```diff
@@ -8,8 +8,10 @@ def reset!
     self.layout = self.class.default_layout
   end
 
+  alias_method :own_extra, :extra
+
   def extra(*)
```
luchotc, line 13:
```suggestion
  def extra
```

PR #173, opened 2020-12-22
Feature add game mode

### LINKED ISSUE #171: Allow gamification to be a user feature flag
Nowadays we have two options in organization config: Gamification enabled or disabled. 
A third option is needed in which the organization delegates on the user if gamification should be enabled or not.

In that case, users will have an option in their profiles allowing them to enable and disable gamification. The first time they are presented with that option, it would be nice to prompt them a message or a small explanation about what that implies. Ideally this configuration should be per organization basis to maintain consistency with the idea that the profile belongs to an organization.

felipecalvo:
## :dart: Goal

Let a user enable gamification for a particular organization. Closes https://github.com/mumuki/mumuki-domain/issues/171.

I believe calling this `Game Mode` is much more user-friendly than talking in terms of`Enable Gamification` or such.

### DIFF
--- a/app/models/user_stats.rb
+++ b/app/models/user_stats.rb
@@ -10,6 +10,10 @@ class UserStats < ApplicationRecord
     self.stats_for(user).exp
   end
 
+  def self.game_mode_enabled_for?(user)
+    self.stats_for(user).game_mode_enabled?
+  end
+
   def add_exp!(points)
     self.exp += points
   end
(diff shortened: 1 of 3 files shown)


luchotc (approved):
Niceee! 

Maybe it's time to add a method `stats_here` for user and delegate behavior to it.
I think that will be cleaner for the future.

Also it would be nice to test that new method :smile:

PR #174, opened 2021-01-18
Feature add uppercase mode

felipecalvo:
## :dart: Goal

Add a `Preferences` object to store user preferences and add the first setting, `uppercase_mode`. Needed for https://github.com/mumuki/mumuki-laboratory/pull/1532.

## :spiral_notepad: Note for reviewers

This is very similar as to how submissions are embedded in assignments, still, please let me know if this is what you were expecting @flbulgarelli as we talked on that other pull request. I'll write some tests in the meantime.

Here is how this design works

```ruby
user = User.last
=> #<User:0x0000557776f0a878
 id: 5,
 uppercase_mode: false>

user.preferences
=> #<Preferences:0x000055d33ece4d20 @uppercase_mode=false>
```

```ruby
user = User.last
=> #<User:0x0000557776f0a878
 id: 5,
 uppercase_mode: false>

user.uppercase_mode=true
=> true

user.preferences
=> #<Preferences:0x000055d33ebe3d60 @uppercase_mode=true>
```

```ruby
user = User.last
=> #<User:0x0000557776f0a878
 id: 5,
 uppercase_mode: false>

preferences = Preferences.new
=> #<Preferences:0x00005606976672d8>

preferences.uppercase_mode = true
=> true

user.preferences = preferences
=> #<Preferences:0x00005606976672d8 @uppercase_mode=true>

user
=> #<User:0x0000557776f0a878
 id: 5,
 uppercase_mode: true>
```

```ruby
user = User.last
=> #<User:0x0000557776f0a878
 id: 5,
 uppercase_mode: true>

user.preferences.uppercase?
=> true
```

### DIFF
--- a/app/models/concerns/with_preferences.rb
+++ b/app/models/concerns/with_preferences.rb
@@ -1,0 +1,7 @@
+module WithPreferences
+  extend ActiveSupport::Concern
+
+  included do
+    composed_of :preferences, mapping: %w(uppercase_mode uppercase_mode), constructor: :from_attributes
+  end
+end
--- a/app/models/preferences.rb
+++ b/app/models/preferences.rb
@@ -1,0 +1,17 @@
+class Preferences
+  include ActiveModel::Model
+
+  def self.attributes
+    [:uppercase_mode]
+  end
+
+  attr_accessor *self.attributes
+
+  def self.from_attributes(*args)
+    new self.attributes.zip(args).to_h
+  end
+
+  def uppercase?
+    uppercase_mode
+  end
+end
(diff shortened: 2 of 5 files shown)


flbulgarelli (approved):
Looks good!

We have already discussed with @felipecalvo about some duplications issues but we will postpone it. He will comment appropriately

felipecalvo:
We decided to keep this as-is but consider that each setting must be added three times right now - one on the list of attributes and two on the mapping. Through a similar solution to the one in `Assignment` `composed_of` `submission`, we can zip the list of attributes with itself, making it so we only have to write each setting once.

This will be implemented once a second setting is added

PR #175, opened 2021-01-18
Feature exam registration process

faloi:
See mumuki/mumuki-laboratory#1541

This adds support at domain-level for exam registrations.


app/models/exam_authorization_request.rb
```diff
@@ -0,0 +1,24 @@
+class ExamAuthorizationRequest < ApplicationRecord
+  belongs_to :exam
+  belongs_to :user
+  belongs_to :organization
+
+  enum status: %i[pending approved rejected]
```
luchotc, line 6: Just nitpicking here, we usually use %i() notation instead of %i[]

app/models/exam_authorization_request.rb
```diff
@@ -0,0 +1,24 @@
...
+  before_save :set_default_status!
+  after_update :notify_user!
+
+  def try_authorize!
+    exam.authorize! user if approved?
+  end
+
+  private
+
+  def set_default_status!
```
luchotc, line 17: As this model is new and we have no records, we could add this default as a database default

app/models/exam_registration.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def ensure_valid_authorization_criterion!
+    authorization_criterion.ensure_valid!
+  end
+
+  def set_default_criterion_type!
+    self.authorization_criterion_type ||= :none
+  end
+
+  def start!(users)
```
luchotc, line 27: Nitpicking again! Here we prefer each(&:notify_user!) syntax

app/models/exam_registration/authorization_criterion.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+end
+
+class ExamRegistration::AuthorizationCriterion::PassedExercises < ExamRegistration::AuthorizationCriterion
+  def valid?
+    value.positive?
+  end
+
+  def enabled_for?(user)
+    user.passed_submissions_count >= value
```
luchotc, line 59: This value should be scoped within the authorization organization

app/models/exercise.rb
```diff
@@ -10,6 +10,7 @@ class Exercise < ApplicationRecord
   include WithNumber,
           WithName,
           WithAssignments,
+          WithProgress,
```
luchotc, line 13: I think this isn't necessary now, because I recall you were adding it in order to use `completed_for?`
If it's in fact necessary, we should discuss another way of including it. Exercise doesn't meet the required interface to be a progress item, as it doesn't have structural_children, for example.

db/migrate/20210114201454_add_exam_registration_to_exam.rb
```diff
@@ -0,0 +1,5 @@
+class AddExamRegistrationToExam < ActiveRecord::Migration[5.1]
+  def change
+    add_reference :exams, :exam_registration, index: true
```
luchotc, line 3: This one should be a many to many relation.
The same exam can be in several registrations

app/models/exam_registration/authorization_criterion.rb
```diff
@@ -0,0 +1,61 @@
...
+  def ensure_valid!
+    raise "Invalid criterion value #{value} for #{type}" unless valid?
+  end
+
+  def process_request!(authorization_request)
+    authorization_request.update! status: authorization_status_for(authorization_request.user)
+  end
+
+  def authorization_status_for(user)
+    enabled_for?(user) ? :approved : :rejected
```
luchotc, line 25: I would change `enabled_for?` for something like `meets_authorization_criteria?`

luchotc (approved):
Just minor comments mostly!
The general implementation looks great! :clap: :100:

app/models/exam_authorization_request.rb
faloi: Ok, I didn't even know about the `%i[...]` syntax, Rubocop suggested it. Perhaps in the future you can include a Rubocop config file to enforce Mumuki style.

app/models/exam_authorization_request.rb
faloi: Great, the default is indeed in the migration, I copy-pasted this from somewhere else and wasn't really sure if it was necessary.

app/models/exam_registration.rb
faloi: But `notify_user!` is not a message of the user here, it's defined on `ExamRegistration` class.

app/models/exercise.rb
faloi: Right, undoing this.

app/models/exam_registration/authorization_criterion.rb
faloi: Ok... let's discuss how to implement it.

db/migrate/20210114201454_add_exam_registration_to_exam.rb
faloi: Done!

app/models/exam_registration/authorization_criterion.rb
faloi: Done!

spec/models/exam_registration_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+    let(:criterion_value) { 2 }
+    let(:exam) { create(:exam, exam_registrations: [registration]) }
+    let!(:auth_requests) do
+      [user, other_user].map { |it| create(:exam_authorization_request, exam: exam, user: it) }
+    end
+
+    before { assignments_for(user, 3).each(&:passed!) }
+    before { registration.process_requests! }
+
+    # TODO: cómo sacar estos reload?
```
faloi, line 36: This was a question for you @luchotc. 😅

PR #176, opened 2021-01-19
Feature prevent manual content

flbulgarelli:
# :dart: Goal

To allow content with `manual_evaluation` to be used in organizations than don't have teachers and thus no one is going to evaluate the submissions. 

# :memo: Details

This PR introduces the `prevent_manual_evaluation_content` flag for organizations that won't manually evaluate submission.  When enabled, exercises with `manual_evaluation` are simply wiped locally.     

This PR relies on `default_scope`. Yes, I know that using it is largely discouraged (eg [here](https://andycroll.com/ruby/dont-use-default-scope/) and [here](https://stackoverflow.com/questions/25087336/why-is-using-the-rails-default-scope-often-recommend-against)). However, I think this is a case where it still applies, and using named scopes would require a huge refactor to the whole platform. 

Actually, I think it is good to completely ignore manual exercises in such contexts, like it were a soft delete. I can't think of a scenario where `unscoped` would be actually necessary, since you don't edit guides from an organization - you are either in a organization-less script or implicitly using `base`.   

# :soon: Future work

This feature may be subsumed by adaptive paths in the future, since it also deals with skipping exercises in some sense. 

This PR is completely backwards compatible.  

# :warning: Warnings

* `base` organization should never have this flag enabled. Otherwise, you won't be able to edit manual evaluation content :scream: 
* If manual evaluation exercises are not in the last position of the guide, numeration will get inconsistent. This PR does not tries to fix that - manual evaluation that needs to be used  in these contexts must place such exercises at last.

### DIFF
--- a/app/models/exercise.rb
+++ b/app/models/exercise.rb
@@ -29,7 +29,7 @@ class Exercise < ApplicationRecord
   defaults { self.submissions_count = 0 }
 
   def self.default_scope
-    where(manual_evaluation: false) if Organization.safe_current&.hide_manual_evaluation_content
+    where(manual_evaluation: false) if Organization.safe_current&.prevent_manual_evaluation_content
   end
 
   alias_method :progress_for, :assignment_for
--- a/app/models/exercise.rb
+++ b/app/models/exercise.rb
@@ -28,6 +28,10 @@ class Exercise < ApplicationRecord
 
   defaults { self.submissions_count = 0 }
 
+  def self.default_scope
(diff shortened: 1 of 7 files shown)


luchotc (approved):
Love the introduction of `safe_current`.
It truly seems a great use case for default_scope. Exercise initialization with manual_evaluation = false is not a problem and it's not likely that we want that unscoped.
The only thing that is going to bother us is when using rails console. We have to be aware of that.

PR #177, opened 2021-01-25
Chore database optimization

flbulgarelli:
# :dart: Goal

To prove new methods that allow database optimization in mumuki-laboratory. 

# :memo: Details

Each commit is enumerated according to its optimization number that uniquely identifies the change across both domain and laboratory repositories. Actual optimization results are shown in laboratory's PR. 

In particular, the following new methods were introduced: 

* `assignments_for`, `statuses_for` at `Guide`, which are the recommended ways now of fetching a bunch of assignments instead of mapping to individual fetches. They mimic the original `WithAssignments` API but in a pluralized fashion.
* `almost_done?` at `Stats`. Side note: this object should be renamed someday to `GuideStats`, `ProgressStats`, `GuideProgressStats` or something like that. Maybe it should be subsumed into indicators @julian-berbel @felipecalvo 

Also, the only truly locally optimized method was `pending_lessons` at chapter.    

# :book: Suggested lecture

If not familiar with, I suggest reading about `includes` and family. Here are some posts: 

* https://scoutapm.com/blog/activerecord-includes-vs-joins-vs-preload-vs-eager_load-when-and-where
* https://medium.com/@tadhao/joins-vs-preload-vs-includes-vs-eager-load-in-rails-5f721c44b3a9

Unfortunately I didn't found good sources about [Preloader](https://www.rubydoc.info/docs/rails/ActiveRecord/Associations/Preloader). I actually found it digging in gists and Stack Overflow

# :eyes:  See 

* https://github.com/mumuki/mumuki-laboratory/pull/1552

### DIFF
--- a/app/models/concerns/with_assignments_batch.rb
+++ b/app/models/concerns/with_assignments_batch.rb
@@ -9,10 +9,12 @@ module WithAssignmentsBatch
 
     return exercises.map { |it| block.call nil, it  } unless user
 
-    exercises = self.exercises
-    ActiveRecord::Associations::Preloader.new.preload(exercises, :assignments, Assignment.where(submitter: user))
+    pairs = exercises.map { |it| [it.id, [nil, it]] }.to_h
+    Assignment.where(submitter: user, exercise: exercises).each do |it|
+      pairs[it.exercise_id][0] = it
+    end
 
-    exercises.map { |it| block.call it.assignments.first, it }
+    pairs.values.map { |assignment, exercise| block.call assignment, exercise }
   end
 
   def statuses_for(user, organization = Organization.current)
--- a/app/models/concerns/with_assignments_batch.rb
+++ b/app/models/concerns/with_assignments_batch.rb
@@ -4,15 +4,15 @@
 module WithAssignmentsBatch
   extend ActiveSupport::Concern
 
-  def find_assignments_for(user, _organization = Organization.current)
+  def find_assignments_for(user, _organization = Organization.current, &block)
+    block = block_given? ? block : lambda { |it, _e| it }
+
+    return exercises.map { |it| block.call nil, it  } unless user
+
     exercises = self.exercises
     ActiveRecord::Associations::Preloader.new.preload(exercises, :assignments, Assignment.where(submitter: user))
 
-    if block_given?
-      exercises.map { |it| yield it.assignments.first, it }
-    else
-      exercises.map { |it| it.assignments.first }
-    end
+    exercises.map { |it| block.call it.assignments.first, it }
   end
 
   def statuses_for(user, organization = Organization.current)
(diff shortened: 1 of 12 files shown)


app/models/concerns/with_assignments_batch.rb
```diff
@@ -0,0 +1,29 @@
+# WithAssignmentsBatch mirrors the WithAssignment mixin
+# but implements operations in batches, so that they outperform
+# their counterparts
+module WithAssignmentsBatch
+  extend ActiveSupport::Concern
+
+  def find_assignments_for(user, _organization = Organization.current)
+    exercises = self.exercises
```
julian-berbel, line 12: doesn't the attribute get cache'd?

julian-berbel (approved):
Looks good! :heart:

app/models/concerns/with_assignments_batch.rb
luchotc: I guess that because of how he uses it, it is preloading it on exercises object and not into guide itself. So it needs to assign it to a variable in order to access through that preload in the last line.

luchotc (approved):
Great :heart: :100:

app/models/concerns/with_assignments_batch.rb
flbulgarelli: Exactly. You need to preload a given instance of the relation object. If the relation instance changes, the preloading gets lost. 

I have actually seen samples that go even further and convert the relation into an array before doing this.

flbulgarelli: Long story short, I was wrong. As @julian-berbel said, the collection proxy get cached, which means that this implementation had side-effects :open_mouth: 

I have re-implemented this method in a less obscure but more algorithmic way, which ensures no side effect is produced. Please review it again :pray:

luchotc (approved):
Definitely we are side-effects free now! Also, it's a bit more readable this way, so win-win.

julian-berbel (approved):

app/models/concerns/with_assignments_batch.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    return exercises.map { |it| block.call nil, it  } unless user
+
+    pairs = exercises.map { |it| [it.id, [nil, it]] }.to_h
+    Assignment.where(submitter: user, exercise: exercises).each do |it|
+      pairs[it.exercise_id][0] = it
+    end
+
+    pairs.values.map { |assignment, exercise| block.call assignment, exercise }
+  end
```
felipecalvo, line 18: I find having both `assignments_for` and `find_assignments_for` a tad confusing. Does `find_assignment_for` have any purpose other than being of use to `statuses_for` and `assignments_for`? Are there any use cases where we would want to call it directly?

I thought of suggesting it should be private but I see you're using it to test on `guide_spec.rb`.

felipecalvo (approved):
This is great! :star_struck:

app/models/concerns/with_assignments_batch.rb
flbulgarelli: I see your point. I simply put it public to mimic the `WithAssignments` mixin, which also has a `find_assignment_for` with the same semantics: _find_ it returning nil if it does not exist, in the same manner as the `find_by` methods. 

It may also be useful for other implementations, but it is mostly for sake of consistency. Both methods may turn private in the future.

PR #179, opened 2021-01-28
Exam registration process UI

faloi:
See mumuki/mumuki-laboratory#1541 and https://github.com/mumuki/mumuki-laboratory/pull/1559

This adds some handy methods that are used from the UI.


luchotc (approved):
Nice :clap: :100:

PR #180, opened 2021-02-01
Chore more db optimizations

flbulgarelli:
# :dart: Goal

To implement some additional method specializations that come with or allow performance improvements.

This PR is fully backwards compatible


app/models/exercise.rb
```diff
@@ -42,7 +42,6 @@ def self.default_scope
 
   randomize(*RANDOMIZED_FIELDS)
   delegate :timed?, to: :navigable_parent
-  delegate :stats_for, to: :guide
```
flbulgarelli, line 45: I should instead keep it with a deprecation notice

PR #182, opened 2021-02-08
Add activity stats to UserStats

faloi:
## :dart: Goal

Add methods to retrieve information needed for mumuki/mumuki-laboratory#1540.

## :memo: Details

This was added to `UserStats` to avoid cluttering (even more!) the `User` model.

### DIFF
--- a/app/models/user_stats.rb
+++ b/app/models/user_stats.rb
@@ -11,11 +11,12 @@ class UserStats < ApplicationRecord
   end
 
   def activity(date_range = nil)
+    date_filter = { submitted_at: date_range }.compact
     {
         exercises: {
             solved_count: organization_exercises
                               .joins(:assignments)
-                              .where(assignments: { top_submission_status: [:passed, :skipped], submitter: user }.merge_if_present(:submitted_at, date_range))
+                              .where(assignments: { top_submission_status: [:passed, :skipped], submitter: user }.merge(date_filter))
                               .count,
             count: organization_exercises.count},
 
@@ -32,17 +33,12 @@ class UserStats < ApplicationRecord
   private
 
   def messages_in_discussions(date_range = nil)
+    date_filter = { date: date_range }.compact
     Message.joins(:discussion)
-        .where({sender: user.uid, discussions: { organization: organization }}.merge_if_present(:date, date_range))
+        .where({sender: user.uid, discussions: { organization: organization }}.merge(date_filter))
   end
 
   def organization_exercises
     organization.book.exercises
-  end
-end
-
-class Hash
-  def merge_if_present(key, value)
-    value ? merge({ key => value }) : self
   end
 end
--- a/app/models/user_stats.rb
+++ b/app/models/user_stats.rb
@@ -10,7 +10,39 @@ class UserStats < ApplicationRecord
     self.stats_for(user).exp
   end
 
+  def activity(date_range = nil)
+    {
+        exercises: {
+            solved_count: organization_exercises
+                              .joins(:assignments)
+                              .where(assignments: { top_submission_status: [:passed, :skipped], submitter: user }.merge_if_present(:submitted_at, date_range))
+                              .count,
+            count: organization_exercises.count},
+
+        messages: {
+            count: messages_in_discussions(date_range).count,
+            approved: messages_in_discussions(date_range).where(approved: true).count}
+    }
+  end
+
   def add_exp!(points)
     self.exp += points
+  end
+
+  private
+
+  def messages_in_discussions(date_range = nil)
+    Message.joins(:discussion)
+        .where({sender: user.uid, discussions: { organization: organization }}.merge_if_present(:date, date_range))
+  end
+
+  def organization_exercises
+    organization.book.exercises
+  end
+end
+
+class Hash
+  def merge_if_present(key, value)
+    value ? merge({ key => value }) : self
   end
 end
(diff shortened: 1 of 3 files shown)


app/models/user_stats.rb
```diff
@@ -10,7 +10,39 @@ def self.exp_for(user)
...
+        .where({sender: user.uid, discussions: { organization: organization }}.merge_if_present(:date, date_range))
+  end
+
+  def organization_exercises
+    organization.book.exercises
+  end
+end
+
+class Hash
+  def merge_if_present(key, value)
```
luchotc, line 45: I think this is a good approach but I'm worried about cases like {do_something_bad: false}, which would be filtered and I don't think we want that :confused:

app/models/user_stats.rb
```diff
@@ -10,7 +10,39 @@ def self.exp_for(user)
...
+
   def add_exp!(points)
     self.exp += points
   end
+
+  private
+
+  def messages_in_discussions(date_range = nil)
+    Message.joins(:discussion)
+        .where({sender: user.uid, discussions: { organization: organization }}.merge_if_present(:date, date_range))
```
luchotc, line 36: I would propose to take advantage of compact and just do:

data_param = {date: date_range}.compact
.....
merge(date_param)

:smile:

app/models/user_stats.rb
```diff
@@ -10,7 +10,39 @@ def self.exp_for(user)
     self.stats_for(user).exp
   end
 
+  def activity(date_range = nil)
+    {
+        exercises: {
+            solved_count: organization_exercises
+                              .joins(:assignments)
+                              .where(assignments: { top_submission_status: [:passed, :skipped], submitter: user }.merge_if_present(:submitted_at, date_range))
```
luchotc, line 18: This is not necessary to be changed, just a reminder to me to create an issue before merging this.
There are several places where we are doing passed? || skipped? or things like that. We should at least define a method in assignment like `solved_statuses` and use that instead.

app/models/user_stats.rb
```diff
@@ -10,7 +10,39 @@ def self.exp_for(user)
...
   end
+
+  private
+
+  def messages_in_discussions(date_range = nil)
+    Message.joins(:discussion)
+        .where({sender: user.uid, discussions: { organization: organization }}.merge_if_present(:date, date_range))
+  end
+
+  def organization_exercises
```
luchotc, line 39: @flbulgarelli accessing directly through guide usages will be faster here, right?
With @julian-berbel we've been thinking that maybe an organization content caché would improve our performance :monocle_face: 

Again, @faloi, this does not need to be changed, we'll discuss it and change it afterwards!

luchotc (approved):
So clean! :100:  Just minor comments.
Kudos for adding those tests :clap:

app/models/user_stats.rb
flbulgarelli: In any case, let's please add extension methods in the appropriate directory in order to keep things organized and avoid problems with rails lazy-loading. 

See https://github.com/mumuki/mumuki-domain/blob/master/lib/mumuki/domain/extensions/hash.rb

app/models/user_stats.rb
flbulgarelli: Yup, that is true. We could add an `organization.exercises` that does the same

app/models/user_stats.rb
flbulgarelli: We also have a lot of code that does the opposite too: `failed_statuses`

app/models/user_stats.rb
```diff
@@ -10,7 +10,39 @@ def self.exp_for(user)
...
+                              .joins(:assignments)
+                              .where(assignments: { top_submission_status: [:passed, :skipped], submitter: user }.merge_if_present(:submitted_at, date_range))
+                              .count,
+            count: organization_exercises.count},
+
+        messages: {
+            count: messages_in_discussions(date_range).count,
+            approved: messages_in_discussions(date_range).where(approved: true).count}
+    }
+  end
```
flbulgarelli, line 26: The whole method looks like expensive in database terms. Not a big issue since this route will not be a bottleneck, but for example instead of querying `messages_in_discussions` twice, we could do it once and filter and count in-memory, or select both `count(approved)` and `count(1)`, or even group-by `approved`.

app/models/user_stats.rb
```diff
@@ -10,7 +10,39 @@ def self.exp_for(user)
...
+
   def add_exp!(points)
     self.exp += points
   end
+
+  private
+
+  def messages_in_discussions(date_range = nil)
```
flbulgarelli, line 34: This is more a though for the future than a suggestion, but we should prefer named args when using optional arguments to make things cleaner. Something like `messages_in_discussions range: a_range`.

luchotc: Agree!

app/models/user_stats.rb
faloi: Nice one. SQL is not my thing... 😅

app/models/user_stats.rb
```diff
@@ -10,7 +10,39 @@ def self.exp_for(user)
...
+
   def add_exp!(points)
     self.exp += points
   end
+
+  private
+
+  def messages_in_discussions_count(date_range = nil)
+    date_filter = { date: date_range }.compact
+    result = Message.joins(:discussion)
```
luchotc, line 35: Kudos for optimizing it so much :smile:

luchotc (approved):
Looks great @faloi!

app/models/user_stats.rb
luchotc: Good news, that method already exists and does what we want :smile:

PR #184, opened 2021-02-17
Chore preferences

felipecalvo:
## :dart: Goal

Make some preferences-related small changes needed by Laboratory.

## :memo: Details

* Adds `uppercase?` as a shorthand for `uppercase_mode`
* Kids organizations are now either `primary` or `kindergarten`. This is so we can auto-uppercase kindergarten organizations instead of relying on custom CSS. All current `kids` organizations will now be `primary`.
* Makes all preferences `false` for incognito user.

## :soon: Future work

I might just merge this other PR https://github.com/mumuki/mumuki-domain/pull/173 (and a few more changes) into this one.

### DIFF
--- a/lib/mumuki/domain/incognito.rb
+++ b/lib/mumuki/domain/incognito.rb
@@ -84,6 +84,20 @@ module Mumuki::Domain
     end
 
     # ========
+    # Preferences
+    # ========
+
+    def preferences
+      Class.new do
+        class << self
+          def method_missing(_)
+            false
+          end
+        end
+      end
+    end
+
+    # ========
     # Visiting
     # ========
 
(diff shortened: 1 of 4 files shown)


lib/mumuki/domain/incognito.rb
```diff
@@ -83,6 +83,20 @@ def has_accepted?(term)
       false
     end
 
+    # ========
+    # Preferences
+    # ========
+
+    def preferences
+      Class.new do
```
flbulgarelli, line 91: :warning: This is creating a new class in each method call, but you could use always the same class and even the same instance.

felipecalvo: Right. Do you mean like this? Or is there a better option?

```suggestion
      @preferences ||= Class.new do
```

flbulgarelli: I was actually thinking of a named module like `module NullPreferences` within incognito

PR #185, opened 2021-02-17
Add currently_in_exam? method for incognito

luchotc:
Needed by https://github.com/mumuki/mumuki-laboratory/pull/1573

Incognito user needs to respond to that method in order to be able to access forum related routes without an error. It isn't actually allowed to access them, but with this one a proper error message is renderer. Tested in labo associated branch


felipecalvo (approved):
Let's go :open_hands:

PR #186, opened 2021-02-22
Use organization exercises method and cache result

luchotc:
Just a minor improvement to @faloi https://github.com/mumuki/mumuki-domain/pull/182
I'm using organization exercises method which access through usages and I'm using memoization to avoid asking for it twice.

### DIFF
--- a/app/models/user_stats.rb
+++ b/app/models/user_stats.rb
@@ -43,6 +43,6 @@ class UserStats < ApplicationRecord
   end
 
   def organization_exercises
-    organization.book.exercises
+    @organization_exercises ||= organization.exercises
   end
 end


app/models/user_stats.rb
```diff
@@ -43,6 +43,6 @@ def messages_in_discussions_count(date_range = nil)
   end
 
   def organization_exercises
-    organization.book.exercises
+    @organization_exercises ||= organization.exercises
```
flbulgarelli, line 46: Why not just caching it in exercises method of organization?

luchotc: Because it's defined with a `has_many` association so it is cached by default.
With the current implementation of this method no memoization is needed at all. I added it to be extra sure in case that implementation changes :stuck_out_tongue:

flbulgarelli: _controversial_

PR #188, opened 2021-02-22
Feature certifications

fedescarpa:


app/models/certificate.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  def locals
+    json = as_json only: [:start_date, :end_date],
+                   include: {
+                     user: { methods: [:formal_first_name, :formal_last_name, :formal_full_name] },
+                     certification: { only: [:title, :description] },
+                     organization: { only: [:name, :display_name] }
+                   }
+    JSON.parse json.to_json, object_class: OpenStruct
```
luchotc, line 26: Maybe we can extract this behavior to a method in Hash, like `to_os`?

app/models/user.rb
```diff
@@ -275,6 +277,26 @@ def current_immersive_context_and_content_at(path_item)
...
+
+  def formal_last_name
+    verified_last_name || last_name
+  end
+
+  def formal_full_name
+    "#{formal_first_name} #{formal_last_name}"
+  end
+
+  def certificates_in(organization)
```
luchotc, line 292: It isn't expensive but nevertheless why aren't we using an AR query for this one too?

db/migrate/20210119174504_create_certifications.rb
```diff
@@ -0,0 +1,12 @@
+class CreateCertifications < ActiveRecord::Migration[5.1]
+  def change
+    create_table :certifications do |t|
+      t.string :title
+      t.string :template_html_erb
+      t.text :description
+      t.references :organization, foreign_key: true
```
luchotc, line 7: We don't usually create an actual FK. An index would be needed though

luchotc, line 7:
```suggestion
      t.references :organization, index: true
```

db/migrate/20210119174835_create_certificates.rb
```diff
@@ -0,0 +1,13 @@
+class CreateCertificates < ActiveRecord::Migration[5.1]
+  def change
+    create_table :certificates do |t|
+      t.references :user, foreign_key: true
```
luchotc, line 4: same for these two

luchotc (approved):
Just minor details, it looks great @fedescarpa :heart: 
Nice tests :100:

app/models/concerns/with_code.rb
```diff
@@ -0,0 +1,19 @@
+module WithCode
```
flbulgarelli, line 1: Since we use the term code also for referring source-code, I'd rename this to `WithUniqueCode` or `WithGeneratedCode` or something like that.

flbulgarelli: Besides of that, kudos for extracting this mixin :muscle:

app/models/certificate.rb
flbulgarelli: This methods already exists. It is `to_struct`: https://github.com/mumuki/mumukit-core/blob/2f822e552f4f284b8a3563283e4b0c00a409d840/lib/mumukit/core/hash.rb#L33.


The new thing here is that conversion is deep, right? In that case we could have `to_deep_struct`. Anyway, I don't recommend using intermediate JSON generation and parsing, because it is unnecessary slow

app/models/certification.rb
```diff
@@ -0,0 +1,42 @@
+class Certification < ApplicationRecord
```
flbulgarelli, line 1: I don't get the distinction between certificate and certification. I understand the first is the token and the second is the process, but they are even synonyms according to some sources: https://www.merriam-webster.com/thesaurus/certification

flbulgarelli: I think I'd prefer the term `Diploma` instead of `Certificate`

app/models/certificate.rb
luchotc: Yes, that's true! `to_deep_struct` seems like a good name!
We could convert the hash recursively instead of doing this. I think this was the more elegant solution but not the fastest one :thinking:

app/models/certification.rb
flbulgarelli: Alternatively, if changing `Certificate => Diploma`  is a too expensive rename, we could instead rename `Certification` to `Accreditation`: https://www.synonym.com/synonyms/accreditation

app/models/certificate.rb
luchotc: image
There is a big difference, so let's please use a custom solution for this one

app/models/certification.rb
luchotc: Yep, we were mislead by spanish translation of those words. It was meant to be the difference between "la certificación" y "los certificados emitidos".
`Accreditation` and `Certificate` seem to be a good fit for what we want.

app/models/user.rb
luchotc: Also it would be great to have Organization.current as default for organization.
Maybe then the method should be called `certificates_in_organization`

app/models/certification.rb
fedescarpa: Reading this...
> Certification generally refers to an earned credential that demonstrates the holder's specialized knowledge, skills, and experience. Certification differs from a certificate program, which is usually an educational offering that confers a document at the program's conclusion.

I think that `Certificate` is a good name and we should rename `Certifications` by `CertificateProgram`

flbulgarelli: Maybe `CertificateProgram` is a good term. But if it the case, I am confused. What do we want to represent here? 

An `Accreditation` and a `CertificateProgram` is not the same, one refers to the precise moment where examination takes places, and the other to a whole educative process of training and maybe examination. What do you want to represent here @fedescarpa with this class? Is this a single object we create _a priori_ when the project starts, or are we going to create multiple instances for it during its life-cycle? Which will be it's `created_at`?

app/models/user.rb
```diff
@@ -275,6 +277,26 @@ def current_immersive_context_and_content_at(path_item)
...
+  def formal_last_name
+    verified_last_name || last_name
+  end
+
+  def formal_full_name
+    "#{formal_first_name} #{formal_last_name}"
+  end
+
+  def certificates_in_organization(organization = Organization.current)
+    certificates.where certificate_program: CertificateProgram.where(organization: organization)
```
luchotc, line 293: I was going to propose to use a join like:
`user.certificates.joins(:certificate_program).where(certificate_programs: {organization: organization })`

but it surprised me that the query you've come up with does a sub-select internally, so I think in this case with such small tables it's pretty much the same :smile:

spec/models/certificate_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'spec_helper'
+
+describe Certificate, organization_workspace: :test do
+
+  let(:certificate) { create :certificate }
+
+  context '#tmplate_locals' do
```
luchotc, line 7:
```suggestion
  context '#template_locals' do
```

luchotc:
Awesome work @fedescarpa :heart:

PR #189, opened 2021-02-22
Feature split progress by organization

julian-berbel:
## :dart: Goal
To split progress by organization! :sparkles: 
## :memo: Details
Assignments will now only be searched for in the requested organization.

The `skip_unique_name_validation` trait in the `another_test_organization` factory would allow the organization object to be created but it would still prevent the organization from being persisted, which would break a test.

As for the rest of the tests, they are expected behavior changes!

## :warning: Dependencies
None.
Yes.
## :soon: Future work
migration, and lots of testing!

### DIFF
--- a/app/models/concerns/with_assignments.rb
+++ b/app/models/concerns/with_assignments.rb
@@ -19,7 +19,6 @@ module WithAssignments
   end
 
   # TODO: When the organization is used in this one, please change guide.pending_exercises
-  # TODO: Please do the same on WithAssignmentsBatch
   def find_assignment_for(user, organization)
     assignments.find_by(submitter: user, organization: organization)
   end
--- a/app/models/concerns/with_assignments.rb
+++ b/app/models/concerns/with_assignments.rb
@@ -20,8 +20,8 @@ module WithAssignments
 
   # TODO: When the organization is used in this one, please change guide.pending_exercises
   # TODO: Please do the same on WithAssignmentsBatch
-  def find_assignment_for(user, _organization)
-    assignments.find_by(submitter: user)
+  def find_assignment_for(user, organization)
+    assignments.find_by(submitter: user, organization: organization)
   end
 
   def status_for(user)
--- a/spec/models/guide_spec.rb
+++ b/spec/models/guide_spec.rb
@@ -112,13 +112,12 @@ describe Guide do
   end
 
   describe '#clear_progress!' do
-    let(:an_exercise) { create(:exercise) }
-    let(:guide) { create(:indexed_guide) }
+    let(:an_exercise) { create(:indexed_exercise) }
+    let(:guide) { an_exercise.guide }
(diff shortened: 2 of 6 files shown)


luchotc (approved):
:heart:

app/models/concerns/with_assignments.rb
```diff
@@ -20,8 +20,8 @@ def has_messages_for?(user)
 
   # TODO: When the organization is used in this one, please change guide.pending_exercises
   # TODO: Please do the same on WithAssignmentsBatch
-  def find_assignment_for(user, _organization)
```
flbulgarelli, line 23: It is time to fix those TO-DOs, isn't it?


I think that everything will break if we don't change `WithBatchAssignments` too.

julian-berbel: Noticed that during the integration test! fixed!

app/models/concerns/with_assignments.rb
```diff
@@ -19,7 +19,6 @@ def has_messages_for?(user)
   end
 
   # TODO: When the organization is used in this one, please change guide.pending_exercises
```
julian-berbel, line 21: This second TODO shall be addressed in another PR!

flbulgarelli: Great!

luchotc: Awesome!

spec/models/guide_spec.rb
```diff
@@ -140,7 +139,7 @@
     end
 
     context 'when progress is in more than one organization' do
-      let(:another_organization) { create(:another_test_organization) }
+      let(:another_organization) { create(:another_test_organization, book: test_organization.book) }
```
felipecalvo, line 142: Should we add the following test here? Is it relevant?

https://github.com/mumuki/mumuki-domain/commit/5e7759a1fc27e2e1025b7f6ac5b3ddef7b354b56#diff-852e850ded5e5f33c7b8f853db8fbcd980b09b1c93978c29c67facfda1925016R90-R93

It verifies that progress is deleted only on the corresponding organization and not in others, which I think we're missing in this PR (and it's the intended behavior... right? :laughing:)

I found this on a 2019 branch back when we first started this

felipecalvo (approved):
:100: :+1:

spec/models/guide_spec.rb
felipecalvo: Though I guess we should also update this in Laboratory then: https://github.com/mumuki/mumuki-laboratory/commit/e046bc14160f2506f60e844c2a049826da7de1d6

julian-berbel: > Though I guess we should also update this in Laboratory then: mumuki/mumuki-laboratory@e046bc1

that method defaults to current organization so it should be alright either way! still, it would be the responsible thing to add that :grin:

PR #190, opened 2021-02-23
Chore course model improvements

flbulgarelli:
## :dart: Goal

Minor improvements to the courses objects

## :memo: Details

* Removed long standing but hardly attribute's presence validation
* Added `canonical_code`

100% backwards compatible

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -35,6 +35,10 @@ class Course < ApplicationRecord
     end
   end
 
+  def canonical_code
+    "#{period}-#{code}".downcase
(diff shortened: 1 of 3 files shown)


app/models/course.rb
```diff
@@ -35,6 +35,10 @@ def invite!(expiration_date)
     end
   end
 
+  def cannonical_code
```
flbulgarelli, line 38: canonical

luchotc (approved):
Great :+1: :clap:

PR #195, opened 2021-03-09
Added approved_at/by to message model

fedescarpa:
## :dart: Goal
Add `approved_at` and `approved_by` fields to `Message` to save moderator who approved it

## :memo: Details
This is convenient to filter and visualize moderators forum activity

## :warning: Dependencies
No one

No problems

## :soon: Future work
For now we maintain approved field, we must to migrate DB and remove that field from model

### DIFF
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -54,9 +54,9 @@ class Message < ApplicationRecord
 
   def toggle_approved!(user)
     if approved?
-      update! approved: false, approved_at: nil, approved_by: nil
+      disapprove!
     else
-      update! approved: true, approved_at: Time.now, approved_by: user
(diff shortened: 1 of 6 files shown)


app/models/message.rb
```diff
@@ -50,14 +52,22 @@ def read!
     update! read: true
   end
 
-  def toggle_approved!
-    toggle! :approved
+  def toggle_approved!(user)
```
flbulgarelli, line 55: Why not having separate methods for this? Could we delegate this to `approve!` and `disapprove!`?

luchotc (approved):
Apart from @flbulgarelli suggestion, it looks great :smile:

PR #196, opened 2021-03-09
Use organization in pending_* methods

luchotc:
## :dart: Goal

Add organization to pending_* related methods


flbulgarelli:
PS: I fixed the duplication warning

PR #197, opened 2021-03-11
Fix moderator access flags

flbulgarelli:
# :dart: Goal

Fix bug after `future?` refactor in #193
 
# :memo: Details

Some tests added for this functionality, which lacked of coverage


luchotc (approved):
:clap:

PR #198, opened 2021-03-11
Feature delete user account

faloi:
## :dart: Goal

Add a temporary delete account token, as needed by #86.

## :soon: Future work

- [x] Temporary token behavior could be moved to a concern/mixin that takes the field name as parameter and generates both the token creation and check methods.

### DIFF
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -98,7 +98,7 @@ class ApplicationRecord < ActiveRecord::Base
 
       define_method("#{field_name}_matches?") do |token|
         actual_token = attribute(token_attribute)
-        actual_token.present? && token == actual_token && attribute(token_date_attribute) >= Time.now
+        actual_token.present? && token == actual_token && attribute(token_date_attribute)&.future?
       end
     end
   end
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -88,13 +88,12 @@ class ApplicationRecord < ActiveRecord::Base
   end
 
   def self.with_temporary_token(field_name, duration = 2.hours)
-    this = self
     class_eval do
       token_attribute = field_name
       token_date_attribute = "#{field_name}_expiration_date"
 
       define_method("generate_#{field_name}!") do
-        update!(token_attribute => this.generate_secure_token, token_date_attribute => duration.from_now)
+        update!(token_attribute => self.class.generate_secure_token, token_date_attribute => duration.from_now)
       end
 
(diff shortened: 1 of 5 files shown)


app/models/application_record.rb
```diff
@@ -87,6 +87,22 @@ def self.aggregate_of(association)
...
   end
 
+  def self.with_temporary_token(field_name, duration = 2.hours)
+    this = self
+    class_eval do
+      token_attribute = field_name
+      token_date_attribute = "#{field_name}_expiration_date"
+
+      define_method("generate_#{field_name}!") do
+        update!(token_attribute => this.generate_secure_token, token_date_attribute => duration.from_now)
```
luchotc, line 97: We usually use self.class idiom instead of saving the class in a variable

app/models/application_record.rb
```diff
@@ -87,6 +87,22 @@ def self.aggregate_of(association)
...
+      end
+
+      define_method("#{field_name}_matches?") do |token|
+        token == attribute(token_attribute) && attribute(token_date_attribute) >= Time.now
```
luchotc, line 101: We should validate that the token is present to avoid possible security issues

luchotc (approved):
Everything else looks great :clap:

app/models/application_record.rb
faloi: But `self` within the closure is the destination object of the reflection, not the module. Any ideas?

app/models/application_record.rb
faloi: I can also add some kind of validation when this mixin is included. Any thoughts of which error should this throw?

app/models/application_record.rb
luchotc: Sorry, I didn't follow you. `self` outside `class_eval` should evaluate to the destination object (User class in this case) and using `self.class` inside that method definition would evaluate to User class as well. And that's OK because the way you've defined `generate_secure_token` would make it available in that class.

Maybe I'm missing your point though :thinking:

app/models/application_record.rb
luchotc: Do you mean something like throwing `NoMethodError` if those attributes are not defined?
I don't see it as a priority. It would happen anyway on invocation and I don't see a real benefit on doing this eagerly

faloi:
@luchotc please review again.

luchotc (approved):
I've added two extra commits to handle an _improbable_ scenario

PR #199, opened 2021-03-23
Chore certificate programs dates

luchotc:
## :memo: Details
There are several little changes in this one.

First of all I'm introducing optional `start_date` and `end_date` for `CertificateProgram`s and I renamed `start_date` and `end_date` of `Certificate`s to `started_at` and `ended_at` because I think it is more semantic this way and to avoid attribute name repetition and make it clearer.
For a `Certificate`, `started_at` and `ended_at` represent the dates in which the `certified student` actually completed the program. 
For a `CertificateProgram`, `start_date` and `end_date` are optional and they only represent the start and end date of the whole program. A program without start_date means that it is ongoing since the moment it was created and without end_date means that it could be ongoing forever. Ideally each program should have a clear start and end, but I'm considering those cases just in case.

I've created a method in `application_record` for easily creating a scope which filters between two optional dates and allow it to receive a param to specify if the filter should be actually applied or not.

So `CertificateProgram.active(false)` won't actually filter anything, which may be useful _in the future_ :smile:.

I've also added the ability for an `organization` to know if there is any `ongoing_certificate_programs?` there. My idea is to show `My certificates` section only if there are ongoing_certificate_programs or the user was certified in that organization in the past. My intuition is that adding a section for something that you can't obtain, like a certificate it's not ideal. To discuss with @aguspina @lauramangifesta and @felipecalvo  

## :warning: Dependencies
None

100%

### DIFF
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -134,6 +134,19 @@ class ApplicationRecord < ActiveRecord::Base
     end
   end
 
+  def self.active_between(start_date_field, end_date_field, **options)
+    define_singleton_method(:active) do |actually_filter=true|
+      if actually_filter
+        self.where("(#{start_date_field} IS NULL OR #{start_date_field} < :now) AND (#{end_date_field} IS NULL OR #{end_date_field} > :now)", now: Time.now)
+      else
+        all
+      end
+    end
+
+    aliased_as = options.delete(:aliased_as)
+    singleton_class.send(:alias_method, aliased_as, :active) if aliased_as
+  end
+
   ## Partially implements resource-hash protocol, by
   ## defining `to_resource_h` and helper methods `resource_fields` and `slice_resource_h`
   ## using the given fields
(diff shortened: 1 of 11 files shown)


flbulgarelli:
> First of all I'm introducing optional start_date and end_date for CertificatePrograms and I renamed start_date and end_date of Certificates to started_at and ended_at because I think it is more semantic this way and to avoid attribute name repetition and make it clearer.

I understand the need of having two different pair of dates, and I am ok with that. However I don't understand how those names are more intention revealing. For me `started_at` and `start_date` mean exactly the same. Is there a convention I am not getting? Could you explain please? Thanks!

app/models/application_record.rb
```diff
@@ -134,6 +134,19 @@ def self.organic_on(*selectors)
     end
   end
 
+  def self.active_between(start_date_field, end_date_field, **options)
```
flbulgarelli, line 137: Nice!

luchotc:
@flbulgarelli Approved this one verbally

PR #200, opened 2021-03-30
Chore exam registration registrees

luchotc:
## :dart: Goal
We needed to actually save users that were registered in an ExamRegistration. The only bond was the Notification model and that was quite brittle.

100%

### DIFF
--- a/app/models/exam_registration.rb
+++ b/app/models/exam_registration.rb
@@ -6,6 +6,7 @@ class ExamRegistration < ApplicationRecord
   has_and_belongs_to_many :exams
   has_many :authorization_requests, class_name: 'ExamAuthorizationRequest'
   has_and_belongs_to_many :registrees, class_name: 'User'
+  has_many :notifications, as: :target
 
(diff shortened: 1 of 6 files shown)


app/models/exam_registration.rb
```diff
@@ -22,8 +23,12 @@ def ensure_valid_authorization_criterion!
...
   end
 
-  def start!(users)
-    users.each &method(:notify_user!)
+  def register!(users)
+    update! registrees: users
+  end
+
+  def start!
+    registrees.each &method(:notify_user!)
```
fedescarpa, line 31: I think that register should call start only for users that you are registering.
Now, if you start an exam registration and later add some users to the registration you will send the notification to all users, newer and olders.

fedescarpa (changes requested):

luchotc:
@fedescarpa Fair enough. It was addressed so please review it again :smile:

fedescarpa (approved):

PR #201, opened 2021-03-31
Chore banned try solve discussions

luchotc:
## :dart: Goal
Try solve user discussions when it is banned from forum

100%


flbulgarelli (commented):
We discussed this with Lucho and we think it would be better to add a `ban!` entry point with this new logic instead of using `after_save`. 

He will evaluate the effort needed to do this refactor, and proceed with it or this PR as-is based on that.

PR #202, opened 2021-04-06
Feature mixed evaluation

flbulgarelli:
# :dart: Goal

To introduce an specialization of manual evaluation that allows to do some automated checks before putting assignment into `manual_evaluation_pending` status. 

# :memo: Details

The mixed mode is activated when content has manual evaluation flag enabled, but also tests and/or expectations. As such, this mode - like with manual evaluation mode - only makes sense for `Problem`s. 

They logic it follows is quite simple: when a solution to a mixed content is submitted, it is evaluated an usual, but then:

* if status is red - `failed`, `errored` and so on - status is kept red
* if status is yellow, status is turn red - `failed` - too
* otherwise, the solution enters in dark blue - `manual_evaluation_pending` - state  

# :camera_flash: Screenshots

Captures taken from laboratory:

### When expectations fail

Screenshot from 2021-04-07 12-26-00

### When expectations pass

Screenshot from 2021-04-07 12-25-35

Mostly backwards compatible. I have removed two very odd, old methods, but they were not used in the platform aside of some tests

### DIFF
--- a/app/models/exercise.rb
+++ b/app/models/exercise.rb
@@ -249,18 +249,6 @@ class Exercise < ApplicationRecord
   private
 
   def evaluation_class
-    if manual_evaluation?
-      manual_evaluation_class
-    else
-      automated_evaluation_class
-    end
-  end
-
-  def manual_evaluation_class
-    Mumuki::Domain::Evaluation::Manual
-  end
-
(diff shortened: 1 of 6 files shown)


luchotc (approved):
Looks great! :clap:

PR #204, opened 2021-04-09
Feature next chapter

julian-berbel:
~This is a work in progress! :warning:~

## :dart: Goal
Making chapters siblings navigable in order to allow suggesting following chapter once one is completed.
## :memo: Details
Right now this has a small side effect changing the chapter section of breadcrumb from something like `Programación Funcional` to `3. Programación Funcional`; it's easily overridable but I figured I'd ask for opinions first

Also, since I'm restoring pending_siblings for chapters through indicators, should I go ahead and restore it for other contents as well?
## :warning: Dependencies
None.
Yes.
## :soon: Future work
None?

### DIFF
--- a/app/models/chapter.rb
+++ b/app/models/chapter.rb
@@ -29,8 +29,4 @@ class Chapter < ApplicationRecord
   def structural_parent
     book
   end
-
-  def pending_siblings_for(user)
-    book.pending_chapters(user)
-  end
 end
--- a/app/models/chapter.rb
+++ b/app/models/chapter.rb
@@ -2,8 +2,8 @@ class Chapter < ApplicationRecord
   include WithStats
   include WithNumber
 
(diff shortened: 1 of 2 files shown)


julian-berbel:
Opening this PR soon to leave record of a problematic behavior I came across.
This `includes` in container.rb:
https://github.com/mumuki/mumuki-domain/blob/6841eb0018b948768d9fe2f6ba0ddb8dd222fed9/app/models/concerns/container.rb#L35-L37

onflicts with `siblings` in `TerminalNavigation`:
https://github.com/mumuki/mumuki-domain/blob/6841eb0018b948768d9fe2f6ba0ddb8dd222fed9/app/models/concerns/navigation/terminal_navigation.rb#L18-L20

since `TerminalNavigation` returns an array, which does not understand method `includes`

PR #205, opened 2021-04-14
Use only lesson exercises in user_stats activity

luchotc:
## :dart: Goal
Only take into account lesson related exercises

## :memo: Details
There is a problem with the current implementation of user_stats activity. It's using `organization.exercises` which is the fastest way of accessing organization exercises but it has the problem that it's taking into account exams and complements as well, because it's being joined through usages.

There are two alternatives:

- The easy one is the one I've implemented. Using `organization.book.exercises`. This one has the problem that the query is heavier. It's going through the entire content tree to get the exercises.
- The correct one would be to implement an organization method that gets book exercises through usages (using parent_item_type = 'Lesson'). I think in fact, that we could just update `exercises` organization method instead of creating a new one. It's being used in places where it makes sense to access only book exercises.

What do you think @julian-berbel @felipecalvo?

100%

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -33,7 +33,7 @@ class Organization < ApplicationRecord
   after_create :reindex_usages!
   after_update :reindex_usages!, if: lambda { |user| user.saved_change_to_book_id? }
 
-  has_many :guides, through: 'usages', source: 'item', source_type: 'Guide'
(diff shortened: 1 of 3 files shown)


felipecalvo:
Well, you made it pretty clear... I guess the second one _is_ the "right" solution (as long as no current use of `Organization::exercises` requires complements and exams), but I don't know if it requires much extra work.

This is the kind of situation where it's very unlikely that we change it or improve it at a later point, I think.

luchotc:
Great :smile:

PR #206, opened 2021-04-15
Feature soft delete forum messages

felipecalvo:
## :dart: Goal

Let forum messages be soft deleted.

## :memo: Details

The following deletion motives are introduced:
* `self_delete`, intended for users who want to delete their own message;
* `inappropriate_content`,
* `shares_correct_solution` and
* `discloses_personal_information`, all three intended for moderators

Besides the deletion motive, we will also be persisting who deleted the message (`deleted_by`) and when (`deleted_at`)

### DIFF
--- a/app/models/concerns/with_soft_deletion.rb
+++ b/app/models/concerns/with_soft_deletion.rb
@@ -9,4 +9,8 @@ module WithSoftDeletion
   def soft_delete!(motive, deleter)
     update! deletion_motive: motive, deleted_by: deleter, deleted_at: Time.now
   end
+
+  def deleted?
+    deleted_at.present?
+  end
 end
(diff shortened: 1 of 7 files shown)


luchotc (approved):
Looks great :heart: 

Maybe `deletion_motive`  seems too specific to be in a `WithSoftDeletion` mixin but when other use case is added it'll be properly extracted so :ok: for me!

felipecalvo:
We should update this, right?

https://github.com/mumuki/mumuki-domain/blob/102ff99d2d9a3652ea9bb9cba7ae4c819396cdc4/app/models/message.rb#L14

There won't be true destroys anymore.

felipecalvo:
Alright, it's finally done. You might want to quickly check the last four commits, but it's in line with what we talked earlier.

PR #207, opened 2021-04-21
Chore retenantize

luchotc:
## :dart: Goal
Allow organizations to retenantize tenantized paths.
This is useful for the `immersive redirection` feature

## :warning: Dependencies
https://github.com/mumuki/mumukit-platform/pull/46

100%

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,3 +12,4 @@ group :test do
 end
 
 gem 'codeclimate-test-reporter', :group => :test, :require => nil
+gem 'mumukit-platform', github: 'mumuki/mumukit-platform', branch: 'chore-untenantize'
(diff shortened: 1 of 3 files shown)


app/models/discussion.rb
```diff
@@ -47,11 +47,15 @@ def try_solve!
   def navigable_content_in(_)
     nil
   end
-  
+
   def target
     self
   end
 
+  def content_used_in?(organization)
```
luchotc, line 55: This was added because testing locally I've realized that immersive redirection for a specific discussion was not working: https://github.com/mumuki/mumuki-laboratory/pull/1609/commits/168a507db3546293f9eecc37b467c67269fceaf8

PR #208, opened 2021-04-21
Chore abbreviated name

luchotc:
## :dart: Goal
To show abbreviated name instead of user full name in forum messages for anonymity reasons

## :memo: Details
You'll notice that in the last test it returns John George D. instead of John George D. F. . This was intentional to reveal as fewer information possible. It could be easily fixed if you prefer the second case, re-using code from `name_initials` method.

100%

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -211,6 +211,10 @@ class User < ApplicationRecord
     name.split.map(&:first).map(&:capitalize).join(' ')
   end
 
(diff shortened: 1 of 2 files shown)


felipecalvo (approved):
I'd rather not have to do this to be honest but I understand the reasoning.

I still believe we could make it opt out at some point, or introduce usernames or something else.

Let's merge it!

PR #212, opened 2021-04-30
Fix exam breaks on next for

julian-berbel:
## :dart: Goal
Fix exams breaking when one of their exercises is asked for next content.

## :memo: Details
When implementing next chapter button I added a flow that would ask a content's parent for the next available content. Since exams didn't understand message `next_for` this caused them to break when this flow came into action.

Setting TerminalNavigation objects as having no next content fixes this problem.

## :warning: Dependencies
None.

Yes.

## :soon: Future work
None.


fedescarpa (approved):
LGTM

PR #213, opened 2021-05-04
Fix summary access problem

julian-berbel:
## :dart: Goal
Make `affable_test_results` work with string-keyed hashes as well.

## :memo: Details
Honestly, this isn't a great solution; It's _very_ problem-specific and I expect a similar issue could arise again. For the time being though, it fixes our issue.

## :warning: Dependencies
None.

Yes.

## :soon: Future work
None.


felipecalvo (approved):
Alright, very good catch! Still hard to know why we hadn't seen this before.

PR #215, opened 2021-05-07
Add clear_progress_for! to user

julian-berbel:
## :dart: Goal
Ease support tasks that require deleting user progress.

## :memo: Details
Adding `clear_progress_for!` to users to allow clearing progress for a user for a specific organization or optionally, all organizations.

## :warning: Dependencies
None.

Yes.

## :soon: Future work
None.

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -743,4 +743,40 @@ describe User, organization_workspace: :test do
       end
     end
   end
+
+  describe '#clear_progress_for!' do
+    let(:user) { create :user }
+
+    let(:organization_1) { create :organization, name: 'orga-1' }
+    let(:organization_2) { create :organization, name: 'orga-2' }
+    let(:exercise_1) { create :indexed_exercise }
+    let(:exercise_2) { create :indexed_exercise }
+
+    before do
+      organization_1.switch!
+      exercise_1.submit_solution!(user).passed!
+      organization_2.switch!
+      exercise_2.submit_solution!(user).passed!
+    end
+
+    it { expect(user.assignments.count).to eq 2 }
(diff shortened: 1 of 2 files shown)


spec/models/user_spec.rb
```diff
@@ -743,4 +743,48 @@
...
+
+    before do
+      organization_1.switch!
+      assignment_1 = exercise_1.submit_solution!(user).tap(&:passed!)
+      create :message, sender: user, assignment: assignment_1, discussion_id: nil
+      organization_2.switch!
+      assignment_2 = exercise_2.submit_solution!(user).tap(&:passed!)
+      create :message, sender: user, assignment: assignment_2, discussion_id: nil
+      create :message, sender: user, assignment: assignment_1, discussion_id: 1
+    end
```
felipecalvo, line 765: I don't think this last message here can actually occur. Discussion messages don't have an assignment, as far as I know; only teacher-to-student messages do (but then they don't have a discussion id).

In any case the behavior is the same, I think.

felipecalvo (approved):
Good job! I think you're covering all the cases here, I hope we're not missing anything.

spec/models/user_spec.rb
julian-berbel: ah, I was not aware of that. I'll do a quick check on it later just in case

julian-berbel: yep, you were absolutely right! as a side note we have 2 messages which have both `assignment: nil` and `discussion_id: nil`... we should probably delete those

felipecalvo: ~We should delete all messages with `discussion:nil` and the whole feature itself~ Yeah, we should

PR #216, opened 2021-05-10
Fix messages count on user stats

felipecalvo:
## :dart: Goal

Properly count messages on UserStats.

## :memo: Details

The `date` field is used on teacher-to-student `Message`s exclusively, but is `nil` on forum messages. Since the query was filtering by `date`, Laboratory's activity view showed `0 messages` on every week as it never found any results.

Also, we now have soft deleted messages, and we shouldn't count those so I added that.


julian-berbel (approved):
:heart:

PR #222, opened 2021-06-29
Chore misc time zone fixes

flbulgarelli:
# :dart: Goal

To normalize the way we use dates, times and time zones in the code. 

# :memo: Details

We have reviewed time-zone related code with @fedescarpa 

To sum up: 

* Prefer `.current` over `.now`, in order to properly deal with current `Time.zone` 
* Prefer `Time` over `DateTime` just for consistency
* Prefer  `in_time_zone` over `to_time` and `to_datetime` for the same reasons as item 1

# :warning: Dependencies

Depends on #226 

This PR should be 100% backwards compatible. 

# :eyes: See also 

* https://thoughtbot.com/blog/its-about-time-zones
* http://winstonyw.com/2014/03/03/time_now_vs_time_zone_now/
* https://jcxia.com/blog/A-DateTime-bug
* https://gist.github.com/pixeltrix/e2298822dd89d854444b

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -14,7 +14,7 @@ class Course < ApplicationRecord
   resource_fields :slug, :shifts, :code, :days, :period, :description, :period_start, :period_end
 
   def current_invitation
-    invitations.where('expiration_date > ?', Time.now).first
+    invitations.where('expiration_date > ?', Time.current).first
   end
 
   def import_from_resource_h!(resource_h)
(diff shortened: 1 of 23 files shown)


app/models/course.rb
```diff
@@ -49,7 +49,7 @@ def infer_period_range!
     period =~ /^(\d{4})?/
     year = $1.to_i
 
-    return nil unless year.between? 2014, (DateTime.now.year + 1)
+    return nil unless year.between? 2014, (Time.current.year + 1)
```
felipecalvo, line 52: What is this 2014 supposed to mean? Pre-mumuki times?

felipecalvo (approved):
Great, think you caught them all again!

app/models/course.rb
flbulgarelli: I guess so!

flbulgarelli: :up: Yes, that is what it means. We used this in recent times to allow inference of course codes and years which where not normalized  - basically because of classroom legacy code.

julian-berbel (approved):
Looks good!

PR #223, opened 2021-07-01
Use updated_at instead of submitted_at on activity date filters

felipecalvo:
## :dart: Goal

Use a more reliable field for showing weekly activity stats.

## :memo: Details

We recently noticed a bunch of exercises do not have `submitted_at` and therefore can't be filtered by date. Using `updated_at` is not ideal, as there are a few cases which update an Assignment but do not correspond with a submission, but it will do until @flbulgarelli fixes or helps us fix that bug.

## :soon: Future work

This PR should be reverted around three months after when the aforementioned bug is fixed. Twelve weeks is the maximum time we show filters for.

### DIFF
--- a/app/models/user_stats.rb
+++ b/app/models/user_stats.rb
@@ -11,7 +11,7 @@ class UserStats < ApplicationRecord
   end
 
   def activity(date_range = nil)
-    date_filter = { submitted_at: date_range }.compact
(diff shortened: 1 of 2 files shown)


julian-berbel (approved):
Sounds good!

PR #226, opened 2021-07-07
Feature introduce stale message

flbulgarelli:
# :dart: Goal

To introduce `stale` message, which allows platform to track messages sent to previous submissions. 

# :memo: Details

This PR has some important side effects: 

1. An `assignment_id` indexed column has been introduced. This also improves join performance when trying to get messages associated to assignments. That is why I am reverting ba73da87fb2c67cd9db4b9da2fce50c2ba9e539d
2. I have somewhat reified the concept of `direct` - rise hand - messages, as opposed to non-direct ones - forum messages. 
3. I have rewritten the `import_from_resource_h` method, since it basically allowed data without validation to be entered through the private queue. 
4. `stale?`, `direct?` and `contextualization`  messages were introduced. They have no direct usage right now, but are exposed for the future.  

Also, I have refactored `visible_messages`, in order to have that concern in the `Message` class. 

This PR is **not** fully backwards compatible, since: 

* Now previous messages will be kept, changing results of the `messages` methods
* Previously created messages will **not** be joined from now on, since only `assignment_id` will be used. Keeping them will require an additional migration

# :soon: Future work

* I think that we should rename many `messages` associations to `direct_messages` to keep things more intention revealing
* We should change `sender` column to a `sender_id`, in order to avoid joining users using a plain uid.

### DIFF
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -8,18 +8,18 @@ class Message < ApplicationRecord
   has_one :exercise, through: :assignment
 
   validates_presence_of :content, :sender
-  validate :contextualized?
+  validate :ensure_contextualized
 
   after_save :update_counters_cache!
 
   markdown_on :content
 
   # Visible messages are those that can be publicly seen
-  # in forums. non-direct messages are naver visible.
+  # in forums. non-direct messages are never visible.
   scope :visible, -> () do
     where.not(deletion_motive: :self_deleted)
       .or(where(deletion_motive: nil))
-      .where(submission_id: nil)
+      .where(assignment_id: nil)
   end
 
   def contextualization
@@ -135,5 +135,9 @@ class Message < ApplicationRecord
 
   def disapprove!
     update! approved: false, approved_at: nil, approved_by: nil
+  end
+
+  def ensure_contextualized
+    errors.add(:base, :not_properly_contextualized) unless contextualized?
   end
 end
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -2,17 +2,47 @@ class Message < ApplicationRecord
   include WithSoftDeletion
 
   belongs_to :discussion, optional: true
-  belongs_to :assignment, foreign_key: :submission_id, primary_key: :submission_id, optional: true
+  belongs_to :assignment, optional: true
   belongs_to :approved_by, class_name: 'User', optional: true
 
   has_one :exercise, through: :assignment
 
   validates_presence_of :content, :sender
-  validates_presence_of :submission_id, :unless => :discussion_id?
+  validate :contextualized?
 
   after_save :update_counters_cache!
 
   markdown_on :content
+
+  # Visible messages are those that can be publicly seen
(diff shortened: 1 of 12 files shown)


app/models/message.rb
```diff
@@ -2,18 +2,48 @@ class Message < ApplicationRecord
...
+      .or(where(deletion_motive: nil))
+      .where(submission_id: nil)
+  end
+
+  def contextualization
+    direct? ? assignment : discussion
+  end
+
+  def contextualized?
+    assignment_id.present? ^ discussion_id.present?
```
flbulgarelli, line 30: I should improve tests for this method.

app/models/message.rb
```diff
@@ -2,18 +2,48 @@ class Message < ApplicationRecord
...
 
   markdown_on :content
 
+  # Visible messages are those that can be publicly seen
+  # in forums. non-direct messages are naver visible.
+  scope :visible, -> () do
+    where.not(deletion_motive: :self_deleted)
+      .or(where(deletion_motive: nil))
+      .where(submission_id: nil)
+  end
```
flbulgarelli, line 23: Please notice I have added an additional `where` here.

app/models/message.rb
```diff
@@ -2,18 +2,48 @@ class Message < ApplicationRecord
...
   belongs_to :discussion, optional: true
-  belongs_to :assignment, foreign_key: :submission_id, primary_key: :submission_id, optional: true
+  belongs_to :assignment, optional: true
   belongs_to :approved_by, class_name: 'User', optional: true
 
   has_one :exercise, through: :assignment
 
   validates_presence_of :content, :sender
-  validates_presence_of :submission_id, :unless => :discussion_id?
+  validate :contextualized?
```
flbulgarelli, line 11: Now the validation is stricter.

julian-berbel (approved):
Love it! this will require a data migration, right?

flbulgarelli:
> Love it! this will require a data migration, right?

Yup. However, as stated here: 

> Previously created messages will not be joined from now on, since only assignment_id will be used. Keeping them will require an additional migration

The migration is not requiered if previous rise hand messages are not important.

app/models/message.rb
flbulgarelli: Done

app/models/message.rb
```diff
@@ -2,18 +2,48 @@ class Message < ApplicationRecord
...
+  def contextualization
+    direct? ? assignment : discussion
+  end
+
+  def contextualized?
+    assignment_id.present? ^ discussion_id.present?
+  end
+
+  # Whether this message is stale, that is,
+  # targets a submission that is the latest one.
```
felipecalvo, line 34: This caught me off guard for a second - shouldn't it be a submission that is **not** the latest one?

flbulgarelli: You are right! I will change it

flbulgarelli: Done

PR #227, opened 2021-07-12
Fix save submission

flbulgarelli:
# :dart: Goal

To ensure that submissions generate a `submitted_at` as approrpiate 

# :memo: Details

With this PR: 

* Questions won't update `submitted_at` anymore
* Playground will update it
* Tries and Problems will continue to update it.  
* Queries on problem will continue to **not** update it

# :warning: Dependencies

* #222 
* #226 

This PR should be backwards compatible, since the previous behavior was never intended. Issue https://github.com/mumuki/mumuki-domain/pull/223 is an example of the consequences of such obscure code.

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -82,7 +82,7 @@ class Assignment < Progress
     end
   end
 
-  def persist_submission!(submission)
+  def save_submission!(submission)
     transaction do
       update! submission_id: submission.id
       update! submitted_at: Time.current
--- a/app/models/exercise/playground.rb
+++ b/app/models/exercise/playground.rb
@@ -3,9 +3,9 @@ class Playground < QueriableChallenge
 
   name_model_as Exercise
 
-  def setup_query_assignment!(assignment, submission)
+  def save_query_submission!(assignment, submission)
     assignment.running!
-    assignment.persist_submission! submission
+    assignment.save_submission! submission
   end
 
   def save_query_results!(assignment)
--- a/app/models/exercise/playground.rb
+++ b/app/models/exercise/playground.rb
@@ -3,8 +3,9 @@ class Playground < QueriableChallenge
 
   name_model_as Exercise
 
-  def setup_query_assignment!(assignment)
+  def setup_query_assignment!(assignment, submission)
     assignment.running!
+    assignment.persist_submission! submission
   end
 
   def save_query_results!(assignment)
--- a/lib/mumuki/domain/factories/exercise_factory.rb
+++ b/lib/mumuki/domain/factories/exercise_factory.rb
@@ -28,12 +28,20 @@ FactoryBot.define do
   end
 
   factory :exercise_base do
+    transient do
+      indexed { false }
+    end
+
     language { guide ? guide.language : create(:language) }
     sequence(:bibliotheca_id) { |n| n }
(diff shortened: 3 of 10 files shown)


app/models/assignment.rb
```diff
@@ -82,7 +82,7 @@ def visible_status
     end
   end
 
-  def persist_submission!(submission)
+  def save_submission!(submission)
```
flbulgarelli, line 85: I renamed `persist_submission!` and `setup_query_assignment!` for consistency, since they actually talked about the same thing as `Submission#save_submission!`

app/models/exercise/playground.rb
```diff
@@ -3,8 +3,9 @@ class Playground < QueriableChallenge
 
   name_model_as Exercise
 
-  def setup_query_assignment!(assignment)
+  def save_query_submission!(assignment, submission)
     assignment.running!
```
flbulgarelli, line 7: :warning: Beware I am changing the original signature of this method.

lib/mumuki/domain/factories/exercise_factory.rb
```diff
@@ -73,7 +81,7 @@
   factory :exercise, parent: :problem
 
   factory :indexed_exercise, parent: :exercise do
-    guide { create(:indexed_guide) }
+    indexed { true }
```
flbulgarelli, line 84: I think it is a good idea to make all exercises factories _indexable_ on demand

spec/models/question_spec.rb
```diff
@@ -15,14 +15,16 @@
...
       it { expect(assignment.messages.count).to eq 1 }
-      it { expect(assignment.submission_id).to_not be nil }
+      it { expect(assignment.submission_id).to be nil }
+      it { expect(assignment.submitted_at).to be nil }
     end
 
     context 'when a question on a previous submission is sent' do
       before do
-        assignment = exercise.submit_solution!(student, content: 'x = 1')
+        assignment = exercise.submit_solution!(student, content: 'x = 1').reload
```
flbulgarelli, line 24: This `reload` call is tricky. It was necessary since otherwise the time was truncated in the database and results were not the same in line 37.

julian-berbel (approved):

spec/models/query_spec.rb
```diff
@@ -12,9 +11,27 @@
...
+
+    context 'a playground exercise' do
+      let(:exercise) { create(:playground) }
+
+      it_behaves_like 'a query submission'
+      it { expect(assignment.submission_id).to_not be nil }
+      it { expect(assignment.submitted_at).to_not be nil }
+    end
+
+    context 'a problem problem' do
```
felipecalvo, line 31: a problem problem :laughing:

lib/mumuki/domain/factories/exercise_factory.rb
felipecalvo: This is great!

felipecalvo (approved):
:raised_hands:

spec/models/query_spec.rb
flbulgarelli: Ups

flbulgarelli: It was very problematic :stuck_out_tongue:

PR #228, opened 2021-07-15
Add verified full name methods for User

felipecalvo:
## :dart: Goal

Let Laboratory show a user's full verified name, if they have one.


flbulgarelli (commented):
Looks good. I remembered there was also a `formal_name` and I put everything on the same file.

felipecalvo:
Onomastic!! :brain: Thanks for the cleanup.

PR #230, opened 2021-07-26
Improve notifications

julian-berbel:
## :dart: Goal
Add fields to allow for custom notifications, and for higher granularity when deciding which emails to ignore for user.

## :memo: Details
Tweaked exam registrations interface to make them easier to use when there's no authorization_request object.

## :warning: Dependencies
None.

This is backwards compatible.

## :soon: Future work
Add tests.

### DIFF
--- a/app/models/discussion.rb
+++ b/app/models/discussion.rb
@@ -183,6 +183,11 @@ class Discussion < ApplicationRecord
     klazz.constantize.find(debatable_id)
   end
 
+  # TODO remove this once discussions generate notifications
+  def subject
+    'discussion'
+  end
+
   private
 
   def messages_by_updated_at(direction = :desc)
--- a/app/models/exam_registration.rb
+++ b/app/models/exam_registration.rb
@@ -13,7 +13,7 @@ class ExamRegistration < ApplicationRecord
 
   before_save :ensure_valid_authorization_criterion!
 
-  delegate :meets_criterion?, :process_request!, to: :authorization_criterion
+  delegate :meets_criterion?, :meets_authorization_criteria?, :process_request!, to: :authorization_criterion
 
   alias_attribute :name, :description
 
--- a/app/models/exam_registration.rb
+++ b/app/models/exam_registration.rb
@@ -13,7 +13,7 @@ class ExamRegistration < ApplicationRecord
 
   before_save :ensure_valid_authorization_criterion!
 
-  delegate :meets_authorization_criteria?, :process_request!, to: :authorization_criterion
+  delegate :meets_criterion?, :process_request!, to: :authorization_criterion
 
   alias_attribute :name, :description
 
@@ -71,6 +71,10 @@ class ExamRegistration < ApplicationRecord
 
   def registered?(user)
     registrees.include? user
+  end
+
+  def meets_criterion?(user)
+    authorization_criterion.meets_criterion?(user, organization)
   end
 
   private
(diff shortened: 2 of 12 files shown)


app/models/notification.rb
```diff
@@ -1,8 +1,17 @@
 class Notification < ApplicationRecord
   belongs_to :user
   belongs_to :organization
-  belongs_to :target, polymorphic: true
+  belongs_to :target, polymorphic: true, optional: true
 
+  enum subject: %i(
+    free_text
```
flbulgarelli, line 7: Perhaps we could use the `custom`,   `announcement` or even `custom_announcement`, since this is is the main motivation of the free text feature.

spec/dummy/db/schema.rb
```diff
@@ -400,6 +400,8 @@
     t.bigint "organization_id"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.integer "subject"
+    t.text "content"
```
flbulgarelli, line 404: If we use one of the latter terms, here we could use `custom_content`, `announcement_content` or `custom_announcement_content`, so that it is clear that this field is thighly coupled to the aforementioned subject type

app/models/discussion.rb
```diff
@@ -183,6 +183,11 @@ def self.debatable_for(klazz, params)
     klazz.constantize.find(debatable_id)
   end
 
+  # TODO remove this once discussions generate notifications
```
flbulgarelli, line 186: 👍

app/models/exam_registration.rb
```diff
@@ -73,6 +73,10 @@ def registered?(user)
     registrees.include? user
   end
 
+  def meets_criterion?(user)
```
flbulgarelli, line 76: Is this method conflicting with delegate call above?

app/models/notification.rb
```diff
@@ -1,8 +1,17 @@
...
 
+  enum subject: %i(
+    custom
+    exam_registration_enabled
+    exam_registration_approved
+    exam_registration_rejected
+    exam_results_available
+    certificate_available)
+
+  after_create :notify_via_email!
```
flbulgarelli, line 14: I am not requesting to change this code, but i find a bit risky to send emails or have such big effects hidden on a create. The same si happening with notifications i think. We should decouple this.

flbulgarelli: It also makes code difficult to test, and to make mistakes when doing manual tasks like migrations.

app/models/user.rb
```diff
@@ -315,6 +315,18 @@ def clear_progress_for!(organization)
     user_stats.where(location).delete_all
   end
 
+  def notify_via_email!(notification)
+    UserMailer.notification(notification).deliver_later unless ignores_notification? notification
+  end
+
+  def ignores_notification?(notification)
+    ignored_notifications.include? notification.subject
```
flbulgarelli, line 323: We could add a test for this

app/models/exam_registration.rb
julian-berbel: well caught!

app/models/user.rb
julian-berbel: will do!

app/models/notification.rb
flbulgarelli: Makes sense.

flbulgarelli (approved):

PR #231, opened 2021-07-28
Add scope for showing unread discussions first

felipecalvo:
## :dart: Goal

Let discussions be ordered by their subscription status: unread first, read second.

## :memo: Details

Needed for https://github.com/mumuki/mumuki-laboratory/pull/1643.


flbulgarelli (commented):
Looks good!

Could we add at least a single test for it? :pray:

Issue #232, opened 2021-08-02
Subscriptions can't be reversed if the message that triggered them is deleted


felipecalvo:
This is not an important bug, but it does cause an inconsistency on the subscriptions/notifications.

If I write a message on a discussion, all subscriptions to that discussion will be marked as unread. However, if I delete my own message, it'll leave no visual trace, yet all subscriptions will stay unread.

It's not as simple as just marking all subscriptions as read again because there could have been _other_ messages that the user hasn't read yet, which triggered the marking as unread before the deleted message did.

If the message gets deleted by a moderator it's a little clearer than if the message is self-deleted, because a notice on what happened is kept on the discussion. So the reason why the user got that notification is clear: there was a message there, it just happened to be deleted by a moderator. It's inconvenient, but it doesn't look like a bug.

Keeping the subscriptions as read or not for moderator-deleted messages would just be a functional decision. On self-deleted messages, though, it is currently a bug.

PR #233, opened 2021-08-02
Fix discussion notifications

felipecalvo:
## :dart: Goal

Fix a bug where notifications for discussions a user has subscribed to were not being shown. Also, add a few subscription tests.

## :memo: Details

When asking for a user's `unread_discussions`, you'd only get those where the user was the initiator. Since that method is used for notifications, there were actually no notifications shown for discussions where the user was not the initiator. (The discussion _would_ show on the profile, though, whether read or unread, as that is done through `watched_discussions`). This would cause a mismatch between the number of notifications and the actual number of unread discussions, as some of them were not being considered (namely, all those that the user didn't create).

In the most extreme cases, such as that of moderators, they would never get any notifications, as they don't create discussions but do get subscribed when they post a message.

I'll add a comment on the affected lines to clarify further.

## :soon: Future work

I found this other bug https://github.com/mumuki/mumuki-domain/issues/232 while fixing this one, but it's less important.

### DIFF
--- a/app/models/concerns/with_discussion_creation/subscription.rb
+++ b/app/models/concerns/with_discussion_creation/subscription.rb
@@ -8,7 +8,7 @@ module WithDiscussionCreation::Subscription
   end
 
   def subscriptions_in_organization
-    subscriptions.joins(:discussion).where(discussion: discussions_in_organization)
+    subscriptions.where(discussion: Organization.current.discussions)
   end
 
   def subscribed_to?(discussion)
--- a/app/models/discussion.rb
+++ b/app/models/discussion.rb
@@ -21,7 +21,7 @@ class Discussion < ApplicationRecord
   scope :no_responsible_moderator, -> { where('responsible_moderator_at < ?', Time.now - MODERATOR_MAX_RESPONSIBLE_TIME)
                                           .or(where(responsible_moderator_at: nil)) }
   scope :pending_review, -> { where(status: :pending_review) }
-  scope :unread_first, -> { (includes(:subscriptions).reorder('subscriptions.read')) }
+  scope :unread_first, -> { includes(:subscriptions).reorder('subscriptions.read', :created_at) }
 
   after_create :subscribe_initiator!
 
--- a/app/models/discussion.rb
+++ b/app/models/discussion.rb
@@ -97,7 +97,7 @@ class Discussion < ApplicationRecord
   end
 
   def mark_subscriptions_as_unread!(user)
-    subscriptions.where.not(user: user).map(&:unread!)
+    subscriptions.where.not(user: user).each(&:unread!)
   end
 
   def submit_message!(message, user)
--- a/app/models/discussion.rb
+++ b/app/models/discussion.rb
@@ -96,7 +96,7 @@ class Discussion < ApplicationRecord
     upvotes.find_by(user: user)
   end
 
-  def unread_subscriptions(user)
+  def mark_subscriptions_as_unread!(user)
     subscriptions.where.not(user: user).map(&:unread!)
   end
 
@@ -104,7 +104,7 @@ class Discussion < ApplicationRecord
     message.merge!(sender: user.uid)
     messages.create(message)
     user.subscribe_to! self
-    unread_subscriptions(user)
+    mark_subscriptions_as_unread!(user)
     no_responsible! if responsible?(user)
   end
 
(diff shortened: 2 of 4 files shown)


app/models/concerns/with_discussion_creation/subscription.rb
```diff
@@ -8,7 +8,7 @@ module WithDiscussionCreation::Subscription
   end
 
   def subscriptions_in_organization
-    subscriptions.joins(:discussion).where(discussion: discussions_in_organization)
+    subscriptions.where(discussion: Organization.current.discussions_in_organization)
```
felipecalvo, line 11: The `joins` is not needed here.

The original `discussions_in_organization` was evaluating to line 8 here, I believe:

https://github.com/mumuki/mumuki-domain/blob/97a0b79d492d873e296b6f34e5e90407f1da3818/app/models/concerns/with_discussion_creation.rb#L1-L10

However, this would only get discussions where the user was the initiator - I'm not entirely sure if this is an automatic thing because of the relation between User and Discussion and the corresponding FK or what. I'm afraid I don't have this very clear, if I'm honest.

By asking for the organization's `discussions_in_organization` (which delegates that to its book), we do get _all_ the discussions, whether the user was the initiator or not:

https://github.com/mumuki/mumuki-domain/blob/97a0b79d492d873e296b6f34e5e90407f1da3818/app/models/book.rb#L20-L22

We can't just move the method from the book to the organization because it's used in Laboratory's controller to list all discussions.

https://github.com/mumuki/mumuki-laboratory/blob/431c50184478841811925075de969cb59b359a8d/app/controllers/discussions_controller.rb#L65-L67

This way, subscriptions do consider all discussions pertaining to the organization.

felipecalvo: Also: the way the whole subscription/notification thing works will have to be refactored, as we've talked about before, but I'm trying to fix this bug as cleanly as possible without breaking anything else.

flbulgarelli: Nice catch. It was completely unnecessary. 

> However, this would only get discussions where the user was the initiator - I'm not entirely sure if this is an automatic thing because of the relation between User and Discussion and the corresponding FK 

Yes, it is because of the FK in `discussions`. The `organic_on` simply creates a methods that introduces the organization as a filter.

flbulgarelli: However, this is still obscure to me, mostly because of names. Why does a subscription have discussions? Isn't subscription a marker of interest over _one_ discussion? I don't get it's semantics. Perhaps as you said, we just have to go on with this and fix the issue, but please add a TODO here since this is something that has no sense for me at all. 

What is worst, we are changing implementation of `subscriptions_in_organization` regarding the discussions involved - we are moving to _initiated_ discussions to _participating_ discussions -, but nothing about this change is being reflected in the signature.

felipecalvo: A subscription does have one discussion only. That `where` is just making a query with a `IN` operator.

```ruby
Subscription.where(discussion_id: [1, 2, 3])
  Subscription Load (1.3ms)  SELECT "subscriptions".*
                             FROM "subscriptions"
                             WHERE "subscriptions"."discussion_id" IN (1, 2, 3)
  ...
```

Just that instead of IDs we're dealing with the discussions themselves, so it's `.where(discussion: ...)`

Regarding the implementation in `subscriptions_in_organization`, I don't really think the signature should change at all. Subscriptions should have always worked with both initiated and participating discussions, in fact, I'd assume they probably did at first and then stopped working without us noticing. I'm positive I'm just fixing a bug, not extending or changing the feature.

app/models/discussion.rb
```diff
@@ -96,15 +96,15 @@ def upvote_for(user)
     upvotes.find_by(user: user)
   end
 
-  def unread_subscriptions(user)
+  def mark_subscriptions_as_unread!(user)
     subscriptions.where.not(user: user).map(&:unread!)
```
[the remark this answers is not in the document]
felipecalvo: I'm not sure, because that predates this PR. I changed it to `each` nevertheless because it's more appropriate.

felipecalvo:
Alright, it's done. I sneaked in one more commit so we can order discussions by creation date. (It's a slightly unrelated thing, see the issue I just linked for more info).

I'm merging as to not delay this any further

PR #235, opened 2021-08-10
Chore notification improvements

julian-berbel:
## :dart: Goal
Add method for showing only current organization notifications.

## :memo: Details
Adding `notifications_in_organization` similarly to messages and discussions methods we already have.

## :warning: Dependencies
None.

Yes.

## :soon: Future work
None.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -62,6 +62,10 @@ class User < ApplicationRecord
     messages.where('assignments.organization': organization)
   end
 
+  def notifications_in_organization(organization = Organization.current)
+    notifications.where(organization: organization)
+  end
+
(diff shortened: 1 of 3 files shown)


app/models/user.rb
```diff
@@ -62,6 +62,10 @@ def messages_in_organization(organization = Organization.current)
     messages.where('assignments.organization': organization)
   end
 
+  def notifications_in_organization(organization = Organization.current)
+    notifications.where(organization: organization)
+  end
```
felipecalvo, line 67: Can't `organic_on` be used as discussions do?

julian-berbel: you are absolutely right!

PR #236, opened 2021-08-12
Add PG lock to Discussion

felipecalvo:
## :dart: Goal

To be used in Laboratory to avoid responsible moderator overlap.

## :memo: Details

See associated Laboratory PR.


felipecalvo:
> Looks good. However, I don't known why we don't add this mixin to `ApplicationRecord` directly.

Moved it to ApplicationRecord and removed all three class-pecific inclusions.

PR #237, opened 2021-08-20
Add restore all indicators api

fedescarpa:
## :dart: Goal
ReadOnly bneeds to recalculate user progress

## :memo: Details
If organization book change all indicators are destroy. When original book is restored progress for user is not recalculated

100%


app/models/user.rb
```diff
@@ -111,6 +111,20 @@ def clear_progress!
     end
   end
 
+  def restore_organization_progress!(organization)
+    assignments_in(organization).each do |assignment|
+      assignment.tap(&:parent).save!
+    end
+  end
+
+  def assignments_in(organization)
```
julian-berbel, line 120: prefer using `organic_on :assignments`
Also, since we're at it, would you remove method `notifications_in_organization` in line 67-69?
That one already has an `organic_on` too

julian-berbel (approved):
LGTM

PR #238, opened 2021-08-23
Fix assignment content updating on queries

julian-berbel:
## :dart: Goal
Stop queries from updating assignment content.

## :memo: Details
Moving content update logic to `PersistentSubmission`.
Also removing some logic I believe to be stale now that we have progress separated by organization?

## :warning: Dependencies
None.

Yes.

## :soon: Future work
None.

### DIFF
--- a/lib/mumuki/domain/submission/base.rb
+++ b/lib/mumuki/domain/submission/base.rb
@@ -52,7 +52,6 @@ class Mumuki::Domain::Submission::Base
   private
 
   def save_submission!(assignment)
-    assignment.content = content
     assignment.recontextualize!
     assignment.save!
   end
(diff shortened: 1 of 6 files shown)


lib/mumuki/domain/submission/base.rb
```diff
@@ -52,8 +52,6 @@ def id
   private
 
   def save_submission!(assignment)
-    assignment.content = content
-    assignment.recontextualize!
     assignment.save!
```
fedescarpa, line 55: Does it make sense for Base to keep saving the assignment or not? I understand that it should only be in PersistentSubmission. I am right? @julian-berbel  @flbulgarelli

fedescarpa (commented):
I only have a question. Everything else LGTM

PR #239, opened 2021-08-30
Add useful exam registration methods

felipecalvo:
## :dart: Goal

Add a few methods needed for the corresponding Laboratory PR.

## :memo: Details

Extract `ended?` for exam registrations as there were repeated uses of `end_time.past?`.

Also, add `multiple_options?` for knowing if there are multiple exams associated with that registration.

Yes

## :bookmark: See also

https://github.com/mumuki/mumuki-laboratory/pull/1667


julian-berbel (approved):
:heart:

PR #240, opened 2021-09-02
Save solution only when content is not empty

julian-berbel:
## :dart: Goal
Prevent losing solution when querying assignment with no content.

## :memo: Details
This is basically a rollback of this PR: https://github.com/mumuki/mumuki-domain/pull/160.
I _do not like this behavior_. I'm working on a better fix for this, but this will do for the time being.

## :warning: Dependencies
None.

Yes.

## :soon: Future work
None.

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -123,7 +123,7 @@ class Assignment < Progress
   end
 
   def content=(content)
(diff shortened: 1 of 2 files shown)


felipecalvo (approved):
Alright, looking forward for the full solution :100:

PR #242, opened 2021-09-07
Add criterion matcher to authorization criterior

fedescarpa:
## :dart: Goal
Add criterion matcher to exam registration authorization criterion

## :memo: Details
With this change will be able to filter students from mongo using the exam authorization criterion model

100%


julian-berbel (approved):

app/models/exam_registration/authorization_criterion.rb
```diff
@@ -62,4 +70,8 @@ def valid?
   def meets_criterion?(user, organization)
     user.passed_submissions_count_in(organization) >= value
   end
+
+  def criterion_matcher
+    { 'stats.passed': { '$gte': value.to_f } }
```
flbulgarelli, line 75: Whoops. How coupled are we to mongo in domain?

PR #243, opened 2021-09-15
Add exam registrations to organization

fedescarpa:


felipecalvo (approved):
Great :raised_hands: 

There are two typos on the first and third commit if you care about it :shushing_face:

PR #244, opened 2021-09-16
Add student-in-exam condition for `discuss_here?`

felipecalvo:
## :dart: Goal

Extend the _can a user discuss here?_ condition to include the user (student) not being currently in exam. This will allow us in Laboratory to not show the menu bar or profile links related to the forum, as they can't discuss during the exam.

## :memo: Details

This doesn't affect teachers or above, as they never formally start an exam, so they never are currently in one.

https://github.com/mumuki/mumuki-domain/blob/2ba7014f833c7e00bf591f2b5d8e78a6afd72102/app/models/exam.rb#L79-L85

Also, added a few tests for both students and teachers.


spec/models/organization_status_spec.rb
```diff
@@ -33,7 +33,10 @@
     let(:exercise2) { create :exercise }
     let(:discussion) { build :discussion, initiator: user }
 
-    before { create :assignment, submitter: user, organization: organization, exercise: exercise1 }
+    before {
+      create :assignment, submitter: user, organization: organization, exercise: exercise1
+      organization.switch!
+    }
```
felipecalvo, line 39: All these tests were using the default `test_organization` instead of the one created specifically for the occasion, because it was never set as current.

app/models/user.rb
```diff
@@ -202,10 +202,6 @@ def completed_containers_with_lookahead(sequence, organization, lookahead: 1)
     sequence[0..count + lookahead - 1]
   end
 
-  # Tells if the given user can discuss in an organization
-  #
-  # This is true only when this organization has the forum enabled and the user
-  # has the discusser pseudo-permission and the discusser is trusted
```
felipecalvo, line 208: I don't think this added much and was outdated anyway, because it now depends on the organization's access mode - and then even on the default mode there's actually more conditions for letting discuss, such as the user not being banned and not being in an exam (which I just added). 

So I removed the comment.

flbulgarelli (approved):

PR #247, opened 2021-09-30
Feature capped exam authorizations

flbulgarelli:
## :dart: Goal

To allow exams to have a cap of students per turn. 

## :memo: Details

This feature introduces: 

* `available_exams` which contains most of the magic here
* `authorize_requests!` and `update_request_by_id!` which may look new, but is actually old code extracted from laboratory

Backward compatible, as long as you run migrations :stuck_out_tongue: 

## :soon: Future work

It would be nice to a more detailed exception than just `Gone` which actually states that the exam is full.

### DIFF
--- a/app/models/exam_registration.rb
+++ b/app/models/exam_registration.rb
@@ -97,15 +97,17 @@ class ExamRegistration < ApplicationRecord
   end
 
   def request_authorization!(user, exam)
-    validate_exam_available! exam
-    authorization_requests.find_or_create_by! user: user do |it|
-      it.assign_attributes organization: organization, exam: exam
+    with_available_exam exam do
+      authorization_requests.find_or_create_by! user: user do |it|
+        it.assign_attributes organization: organization, exam: exam
+      end
     end
   end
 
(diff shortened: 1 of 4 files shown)


flbulgarelli:
@Gustrucco @lauramangifesta

app/models/exam_registration.rb
```diff
@@ -84,9 +96,34 @@ def ended?
     end_time.past?
   end
 
+  def request_authorization!(user, exam)
+    validate_exam_available! exam
```
flbulgarelli, line 100: A test should be added for this validation

app/models/exam_registration.rb
```diff
@@ -84,9 +96,34 @@ def ended?
...
+    authorization_requests.find_or_create_by! user: user do |it|
+      it.assign_attributes organization: organization, exam: exam
+    end
+  end
+
+  def update_authorization_request_by_id!(request_id, exam)
+    validate_exam_available! exam
```
flbulgarelli, line 107: Same comment as before

PR #248, opened 2021-10-07
Chore reference message sender via - Stage 1

julian-berbel:
## :dart: Goal
Stop referencing user via uid in messages.

## :memo: Details
We resolved splitting migration into 2 stages:
We'll do a schema migration only adding the new sender_id column, and do data migration live, and we'll do a second schema migration removing the sender column at a later version. 

This needs to be run between both migrations:

```ruby
Message.distinct.pluck(:sender).each do |sender|
  Message.where('sender = ?', sender).update_all sender_id: User.locate!(sender).id rescue nil
end
```
## :warning: Dependencies
None.

This requires some updates on dependent gems too.

## :soon: Future work
None.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -19,7 +19,8 @@ class User < ApplicationRecord
   has_many :assignments, foreign_key: :submitter_id
   has_many :indicators
   has_many :user_stats, class_name: 'UserStats'
-  has_many :messages, -> { order(created_at: :desc) }, through: :assignments
+  has_many :forum_messages, -> { where.not(discussion_id: nil)  }, class_name: 'Message', foreign_key: :sender_id
+  has_many :direct_messages, -> { order(created_at: :desc) }, class_name: 'Message', source: :messages, through: :assignments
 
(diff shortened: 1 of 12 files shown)


flbulgarelli:
Looks awesome!!!

app/models/user.rb
```diff
@@ -336,6 +336,10 @@ def ignores_notification?(notification)
     ignored_notifications.include? notification.subject
   end
 
+  def forum_messages
+    Message.where(sender: self).or(Message.where('sender = ?', uid)).where.not(discussion_id: nil)
+  end
```
flbulgarelli, line 341: Awesome! I guess this is the line that will change in the next stage/version, right?

flbulgarelli: We may turn into a plain has_manny in the future

julian-berbel: next stage is basically a revert of the last commit!

julian-berbel: > We may turn into a plain has_manny in the future

yeah, that's what it'll be after the revert :laughing:

PR #250, opened 2021-10-15
Feature detach without make ex student

fedescarpa:


app/models/user.rb
```diff
@@ -157,8 +157,8 @@ def attach!(role, course)
     save_and_notify!
   end
 
-  def detach!(role, course)
-    make_ex_student_of! course.slug if student_of?(course.slug) && solved_any_exercises?(course.organization)
+  def detach!(role, course, make_ex_student = true)
```
flbulgarelli, line 160: My two cents: when using optional boolean params, I highly recommend using named params and false by default, in order to be consistent with the codebase.   

That way, this signature should be: `detach!(role, course, deep: false)` - or `hard`, if you prefer.

PR #251, opened 2021-10-15
Feature async massive notifications

fedescarpa:
## :dart: Goal
Add Massive Job

## :warning: Dependencies
* https://github.com/mumuki/mumukit-nuntius/pull/25

100%

### DIFF
--- a/app/models/custom_notification.rb
+++ b/app/models/custom_notification.rb
@@ -3,42 +3,29 @@ class CustomNotification < ApplicationRecord
   belongs_to :organization
 
   has_and_belongs_to_many :users
+  has_one :massive_job, as: :target
   has_many :notifications, as: :target
 
   validates :title, :body_html, presence: true
 
-  def notify_users_to_add!(uids)
-    uids.each do |uid|
-      Mumukit::Nuntius.notify_job! 'CustomNotificationUserAdded', custom_notification_id: id, uid: uid
-    end
+  alias_attribute :description, :title
+
+  def processed?(user)
+    users.include? user
   end
 
-  def add_user_and_notify_via_email!(uid)
-    add_user! User.locate! uid
-  end
-
-  private
-
-  def add_user!(user)
-    return if users.include? user
+  def process!(user)
     transaction do
       users << user
       Notification.create_and_notify_via_email!(organization: organization, user: user, target: self)
     end
   end
 
+  private
+
   class << self
     def subject
       :custom_notification
     end
-
-    def add_user_and_notify_via_email!(custom_notification_id, uid)
-      find(custom_notification_id).add_user_and_notify_via_email!(uid)
-    end
-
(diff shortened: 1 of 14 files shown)


app/models/custom_notification.rb
```diff
@@ -0,0 +1,31 @@
+class CustomNotification < ApplicationRecord
+
+  belongs_to :organization
+
+  has_and_belongs_to_many :users
```
julian-berbel, line 5: I'm not sure I understand this. Isn't the one that represents massivity `MassiveJob`? shouldn't this relationship be there?

app/models/custom_notification.rb
```diff
@@ -0,0 +1,31 @@
...
+  has_one :massive_job, as: :target
+  has_many :notifications, as: :target
```
julian-berbel, line 7: do we use these 2 other associations for anything?

app/models/massive_job.rb
```diff
@@ -0,0 +1,94 @@
...
+
+  has_many :massive_job_failed_items
+
+  delegate :organization, :description, to: :target
+  delegate :uid, :email, :formal_full_name, to: :user
+
+  alias_attribute :failed_items, :massive_job_failed_items
+
+  def friendly
+    description
```
julian-berbel, line 14: Prefer alias_method

app/models/massive_job.rb
```diff
@@ -0,0 +1,94 @@
...
+  end
+
+  def status
+    return :pending if total.zero?
+    return :passed if processed_count == total_count
+    return :failed if total == total_count
+    :processing
+  end
+
+  def icon
```
julian-berbel, line 47: this should probably be in a view helper

app/models/custom_notification.rb
fedescarpa: DONE!

app/models/custom_notification.rb
fedescarpa: DONE!

app/models/massive_job.rb
fedescarpa: DONE!

app/models/massive_job.rb
```diff
@@ -0,0 +1,90 @@
...
+  def increment_failed_count!(uid, error)
+    transaction do
+      increment! :failed_count
+      failed_items.create! uid: uid, message: error.message, stacktrace: error.full_message(highlight: false, order: :top)
+    end
+  end
+
+  class << self
+    def process!(massive_job_id, uid)
+      MassiveJob.find(massive_job_id).process!(uid)
```
julian-berbel, line 83: `self` is MassiveJob here, so explicit reference is not needed.

julian-berbel (approved):
Just nitpicking, this looks good!

PR #256, opened 2021-10-18
Fix in preparation mode for ex students work like organization disabled

fedescarpa:
## :dart: Goal
@NadiaFinzi here is the fix for the behavior that you expect for ex students when organization is in preparation

## :memo: Details
Behavior for ex students is the same for in preparation and disabled organization status.
They only have access to profile and faqs.

100%


julian-berbel (approved):
Looks good

PR #259, opened 2021-10-31
Add email validator dependency

flbulgarelli:
## :dart: Goal

~To make some student identifying data more consistent.~ 
To allow classroom to validate student identifying data. 

## :memo: Details

~Email is guaranteed to be valid as long as it is non-empty.  Also, first_name is always validated on update.~
~:warning: I am not sure if there could be any issue with social accounts when they don't provide an email and then you try to set it on first login. Will it work? @felipecalvo~ 

I have just added a dependency on `email_validator` The rest will be done here https://github.com/mumuki/mumuki-classroom-api/pull/242

This PR is backwards compatible.

### DIFF
--- a/lib/mumuki/domain.rb
+++ b/lib/mumuki/domain.rb
@@ -1,4 +1,5 @@
-require "mumuki/domain/engine"
+require 'email_validator/strict'
+require 'mumuki/domain/engine'
 
 require 'mumukit/core'
 require 'mumukit/core/activemodel'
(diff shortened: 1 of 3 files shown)


julian-berbel (approved):
I had some doubts with the `validate`s, but you already explained it on meet, so looks good to me!

julian-berbel (approved):
Looks good!

